PR #16, opened 2020-03-23
feat: Add checkbox component

wmahad:


src/components/checkbox.js
```diff
@@ -0,0 +1,84 @@
+(() => ({
+  name: 'Checkbox',
+  type: 'CHECKBOX',
```
chrisobdam: Shall we use the type convention for this, and make type `FORM_COMPONENT`

src/components/dataContainer.js
```diff
@@ -21,6 +21,7 @@
     'ROW',
     'CONTAINER_COMPONENT',
     'CONTENT_COMPONENT',
+    'CHECKBOX',
```
chrisobdam: If we follow the convention, you should only add `FORM_COMPONENT` here

PR #19, opened 2020-03-24
Feat/add datepicker ta 1357

jeroenreijs:


src/components/dateTimePicker.js
```diff
@@ -0,0 +1,133 @@
...
+  allowedTypes: [],
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const {
+      label,
+      defaultValue,
+      required,
+      disabled,
+      error,
+      placeholder,
```
vdberganton: I think we can remove the placeholder for this component? 

Don't have a way of showing it I think

jeroenreijs: Yes you can, by emptying the value manually. 
image

PR #27, opened 2020-03-25
feat: add radio group component

wmahad:


vdberganton:
I'm unable to set a default value in the runtime, so not able to have a selected radio button.

src/components/radio.js
```diff
@@ -0,0 +1,136 @@
...
+    );
+
+    const radioData = (radioOptions || '').split('\n');
+    let Radios = radioData.map(option => renderRadio(option, option));
+
+    if (optionType === 'data') {
+      Radios = renderRadio('value', 'Placeholder');
+      if (!isDev) {
+        Radios = (
+          <GetAll modelId={model} skip={0} take={50}>
```
chrisobdam: Shall we add a filter component option, and apply it here?

PR #55, opened 2020-04-01
Feat/add paper and chip

chrisobdam:
@jeroenreijs Can you have a look at my chip color, doesn't seem to work


jeroenreijs:
Since you are only using one override, you can use your custom className, in this case `chip`.
The root key can be removed from the styles object (in which the selector to the material ui class is not correct, '.&M' should be '')

But you can do it with just using this:

src/components/chip.js
```diff
@@ -0,0 +1,38 @@
+(() => ({
+  name: 'Chip',
+  type: 'CONTENT_COMPONENT',
+  allowedTypes: [],
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const { Chip } = window.MaterialUI.Core;
+    const { label, disabled, variant } = options;
+    const isDev = B.env === 'dev';
+    const chip = (
```
jeroenreijs, line 10: It is recommended to capitalize React component names

vdberganton:
The outline option does not work and still shows the background
We don't have a proper icon yet for these components.

PR #57, opened 2020-04-05
Add Icon component

chrisobdam:


src/components/icon.js
```diff
@@ -0,0 +1,40 @@
...
+  icon: 'ContainerIcon',
+  category: 'LAYOUT',
+  type: 'CONTENT_COMPONENT',
+  allowedTypes: [],
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const { Icons } = window.MaterialUI;
+    const isDev = B.env === 'dev';
+    const { icon } = options;
+    const IconComponent = React.createElement(Icons[icon], {
```
jeroenreijs, line 12: This will break, because default value is `None` which does not exist as a Material Icon.
You could either render something else instead of an icon, or remove the `None` option and set an existing icon as default. Why would anyone drag an icon on the canvas and then not pick an icon?

src/components/icon.js
```diff
@@ -0,0 +1,40 @@
+(() => ({
+  name: 'Icon',
+  icon: 'ContainerIcon',
+  category: 'LAYOUT',
```
jeroenreijs, line 4: Why is the icon in the category `LAYOUT`?

src/components/icon.js
```diff
@@ -0,0 +1,40 @@
...
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const { Icons } = window.MaterialUI;
+    const isDev = B.env === 'dev';
+    const { icon } = options;
+    const IconComponent = React.createElement(Icons[icon], {
+      className: classes.root,
+    });
+
+    return isDev ? <div>{IconComponent}</div> : IconComponent;
```
jeroenreijs, line 16: This `<div>` which is needed on the canvas for the drag and drop, has display `block` as default, which means that you can't have to icons next to eachother. In runtime however, this will be just an svg which has display `inline` by default, but an override to `inline-block` by material. So I suggest adding a class to the wrapping DIV on the canvas which has also display `inline-block`.

src/prefabs/icon.js
```diff
@@ -0,0 +1,1287 @@
...
+        {
+          label: 'Icon',
+          key: 'icon',
+          value: 'None',
+          type: 'CUSTOM',
+          configuration: {
+            as: 'DROPDOWN',
+            dataType: 'string',
+            allowedInput: [
+              {
```
jeroenreijs, line 18: I have my doubts of this value. Do our users need a value `None` when they dragged an icon on the canvas? Why did they drag in the first place? And what do they get?

src/prefabs/icon.js
```diff
@@ -0,0 +1,1287 @@
...
+                name: '',
+                value: 'None',
+              },
+              {
+                name: 'FileCopy',
```
jeroenreijs, line 23: This should be placed alphabetically in the list.

PR #59, opened 2020-04-06
feat: add material autocomplete component

jeroenreijs:


vdberganton:
- In runtime the user input is not displayed, it seems to keep rendering the input. 
- The multiple values does not seem to work when trying to add more than 2 items
- It is unclear what the free solo option does
- When there is one chip with multiple value in the autocomplete the placeholder is displayed

On a more positive note: It does work and I can use this to fill in multiple values to a has many property.

PR #61, opened 2020-04-06
add email input

vdberganton:


chrisobdam:
Would it work to add client side validations?

PR #75, opened 2020-04-10
button can now be an icon and trigger action

vdberganton:


src/components/button.js
```diff
@@ -6,53 +6,106 @@
...
+              event.preventDefault();
+              if (!isDev && !loading && linkType === 'action') callAction();
+            };
+            const actionClickHandler = isAction && { onClick: onClickAction };
+            return variant === 'icon' ? (
+              <IconButton {...iconButtonProps} {...actionClickHandler}>
+                {loading
+                  ? Loader
+                  : React.createElement(
+                      Icons[icon === 'None' ? 'Error' : icon],
```
jeroenreijs: The default (when not loading) is error. This means that the button is never going to get a success state? It always is meant to redirect? Why is this only in the icon button?

vdberganton: That is the default Icon when there is no icon selected on the Icon dropdown option.

Seeing this component supports both Button and Icon we need to support the option "None"

vdberganton: Regarding the success state, you're right that hasn't been added

PR #77, opened 2020-04-15
Feat/add tab component ta 1360

jeroenreijs:


src/prefabs/tabs.js
```diff
@@ -0,0 +1,1397 @@
...
+  icon: 'TabGroupIcon',
+  category: 'CONTENT',
+  structure: [
+    {
+      name: 'Tabs',
+      options: [
+        {
+          label: 'Selected tab index',
+          key: 'defaultValue',
+          value: '0',
```
vdberganton, line 12: One small thing, I think for no-coders it makes more sense if we start the active tabs from number 1.

src/components/tabs.js
```diff
@@ -0,0 +1,119 @@
...
+  name: 'Tabs',
+  type: 'CONTENT_COMPONENT',
+  allowedTypes: ['TAB_COMPONENT'],
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const { Tabs, Tab } = window.MaterialUI.Core;
+    const { Icons } = window.MaterialUI;
+    const { Children, env } = B;
+    const {
+      defaultValue,
```
vdberganton, line 11: Can we add a check that the defaultValue actually exists? 

Now it is possible to fill in the value 1000 and it will display nothing because there are only 2 tabs for example.

Make a fallback that it shows the latest tab that exists

PR #81, opened 2020-04-16
Add recaptcha component

stephankaag:
Currently, there is no support yet for serverside verification.

### DIFF
--- a/src/components/recaptcha.js
+++ b/src/components/recaptcha.js
@@ -1,0 +1,90 @@
+(() => ({
+  name: 'Recaptcha',
+  type: 'CONTENT_COMPONENT',
+  allowedTypes: [],
+  orientation: 'HORIZONTAL',
+  jsx: (
+    <div className={classes.root}>
+      {(() => {
+        const isDev = B.env == 'dev';
+        const isReady = () =>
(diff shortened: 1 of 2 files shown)


stephankaag:
Any feedback? :-)

src/components/recaptcha.js
```diff
@@ -0,0 +1,91 @@
...
+            'https://www.google.com/recaptcha/api.js?render=explicit';
+          script.async = true;
+          script.defer = true;
+
+          document.body.appendChild(script);
+
+          return () => {
+            document.body.removeChild(script);
+          };
+        }, []);
```
vdberganton, line 59: This is a very hackie way of using a script for this component and not something that we want to support.

stephankaag: Do you have a recommended alternative way to include a script tag in a component?

vdberganton: No I believe this is currently the only solution for using a script in a component. 

It is also possible to add this to a bundle but we don't want to add this as another dependency to our default bundle. 

We will have to figure out a proper solution to be able to support this.

vdberganton:
Closing this now while working towards a solution for adding dependencies

PR #91, opened 2020-04-21
Fix/add show option to panel prefab

chielwester:


src/prefabs/panel.js
```diff
@@ -78,6 +78,7 @@
             {
               name: 'Column',
               options: [
+                { type: 'TOGGLE', label: 'Show', key: 'show', value: true },
```
vdberganton, line 81: We are replacing the panel component with the extension panel

PR #96, opened 2020-04-22
Add redirect endpoint option to form to redirect after successful submit

chielwester:
Redirects to endpoint after successful form submit. Uses the 'useEndpoint' helper which is available on edge right now


src/components/form.js
```diff
@@ -15,18 +15,26 @@
...
+        const redirectTo =
+          B.env === 'prod' && redirect && B.useEndpoint(redirect);
 
         return (
           <Action actionId={actionId}>
             {(callAction, { data, loading, error }) => (
               <>
+                {(() => {
+                  if (data && redirectTo) {
+                    window.location.hash = redirectTo;
```
jeroenreijs, line 35: Please use `useHistory` instead of window.location, like explained here: https://github.com/bettyblocks/cli/wiki/Component-Helpers#usehistory and here https://github.com/bettyblocks/cli/wiki/Component-Helpers#buseendpoint

PR #106, opened 2020-04-23
Disable submit button when form submit is loading and display loader …

chielwester:
…icon


src/components/button.js
```diff
@@ -27,8 +27,10 @@
     const isDev = B.env === 'dev';
     const isAction = linkType === 'action';
 
+    const isDisabled = parent && parent.loading ? parent.loading : disabled;
```
vdberganton, line 30: This won't work if the button has a column as parent for example.

PR #107, opened 2020-04-23
Disable submit button when form submit is loading and display loader …

chielwester:


vdberganton:
This will not work if the form has nested columns in it

PR #112, opened 2020-04-28
[Button component] include a none option in the prefab.

bryanoemar:
We will need this when we want to execute a component interaction (javascript).


vdberganton:
Hi, thanks for bringing this to our attention.

We have found a different way of resolving this without the need of a none option. 

This issue will soon be resolved so I will close this PR

PR #113, opened 2020-04-28
Bettyblocks/feat/modal component

bryanoemar:
Simple modal which will expose two functions the 'handleOpenModal' and 'handleCloseModal'. 
With these functions the modal can be controlled.

### DIFF
--- a/src/components/modal.js
+++ b/src/components/modal.js
@@ -6,7 +6,7 @@
   jsx: (
     <div className={B.env === 'dev' ? classes.panel : ''}>
       {(() => {
-        const { env, Children } = B;
+        const { env } = B;
         const { Modal } = window.MaterialUI.Core;
 
         const isEmpty = children.length === 0;
--- a/src/components/modal.js
+++ b/src/components/modal.js
@@ -4,7 +4,7 @@
   allowedTypes: ['BODY_COMPONENT', 'CONTENT_COMPONENT'],
   orientation: 'HORIZONTAL',
   jsx: (
-    <div className={classes.panel}>
+    <div className={B.env === "dev" ? classes.panel : ""}>
       {(() => {
         const { env, Children } = B
         const { Modal } = window.MaterialUI.Core
@@ -43,7 +43,7 @@
                     aria-labelledby="simple-modal-title"
                     aria-describedby="simple-modal-description"
               > 
-                    <B.Children><div style={{background: "white"}}>{children}</div></B.Children>
+                    <Children><div style={{background: "white", width: "50%", margin: "0 auto", padding: "20px", overflowWrap: "break-word"}}>{children}</div></Children>
               </Modal>
             </div>
           </div>
--- a/src/components/modal.js
+++ b/src/components/modal.js
@@ -4,54 +4,57 @@
   allowedTypes: ['BODY_COMPONENT', 'CONTENT_COMPONENT'],
   orientation: 'HORIZONTAL',
   jsx: (
-    <div className={B.env === "dev" ? classes.panel : ""}>
+    <div className={B.env === 'dev' ? classes.panel : ''}>
       {(() => {
-        const { env, Children } = B
-        const { Modal } = window.MaterialUI.Core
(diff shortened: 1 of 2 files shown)


src/components/modal.js
```diff
@@ -0,0 +1,166 @@
+(() => ({
+  name: 'Modal',
+  type: 'CONTAINER_COMPONENT',
+  allowedTypes: ['BODY_COMPONENT', 'CONTENT_COMPONENT'],
+  orientation: 'HORIZONTAL',
+  jsx: (
+    <div className={B.env === 'dev' ? classes.panel : ''}>
```
jeroenreijs, line 7: Why do we always have a wrapper DIV? And why is the className `panel` only used in the page builder, not in runtime? Because there is now a lot of styling attached that has no purpose.

src/components/modal.js
```diff
@@ -0,0 +1,166 @@
...
+              ''
+            )}
+
+            <Modal
+              open={open}
+              onClose={() => setOpen(false)}
+              aria-labelledby="simple-modal-title"
+              aria-describedby="simple-modal-description"
+            >
+              <div className={classes.modal}>{children}</div>
```
jeroenreijs, line 54: This component is always shown on top of the window. I'm not sure if that is what we want.

src/components/modal.js
```diff
@@ -0,0 +1,166 @@
...
+          B.defineFunction('handleCloseModal', e => {
+            e.preventDefault();
+
+            setOpen(false);
+          });
+        }, []);
+
+        return (
+          <div>
+            {isDev && options.isVisibleInDev ? (
```
jeroenreijs, line 34: This `invisibleInDev` option hides everything from here, but it doesn't hide the styling that comes with the className `panel` in the wrapping DIV, so you will still see something in the pagebuilder.

src/components/modal.js
```diff
@@ -0,0 +1,166 @@
...
+
+          B.defineFunction('handleCloseModal', e => {
+            e.preventDefault();
+
+            setOpen(false);
+          });
+        }, []);
+
+        return (
+          <div>
```
jeroenreijs, line 33: What is the use of this extra DIV?

src/components/modal.js
jeroenreijs: Remember we only need this extra DIV in the pagebuilder for having the drag and drop events attached.

src/components/modal.js
```diff
@@ -0,0 +1,166 @@
...
+      })()}
+    </div>
+  ),
+  styles: B => theme => {
+    const style = new B.Styling(theme);
+    const getSpacing = (idx, device = 'Mobile') =>
+      idx === '0' ? '0rem' : style.getSpacing(idx, device);
+    return {
+      modal: {
+        background: 'white',
```
jeroenreijs, line 67: Since you are only setting the color, better use `backgroundColor`.

src/components/modal.js
```diff
@@ -0,0 +1,166 @@
...
+        width: '50%',
+        margin: '0 auto',
+        padding: '20px',
```
jeroenreijs, line 70: We prefer using `rem` units instead of `px`

bryanoemar:
Hi @jeroenreijs 

Thanks, for the tips! I'll be looking into that :)

Quick question, I also heard that it's better to use the Dialog component instead of the Modal component. What is the best way to proceed? Simply add those changes in this branch or close this pull request and creating a new one?

src/components/modal.js
bryanoemar: No specific reason. I'm copy/pasting and learning along the way, so thanks for the tips 👍

jeroenreijs:
If you are about to change a lot, then it's maybe better to close this one and create a new one later. You are already behind on the current acceptance, so you can start from there then.
Also, the component interactions will work when merging acceptance.

bryanoemar:
Oki, I'll do just that and create a new merge request. 

Thanks for the help.

PR #118, opened 2020-04-28
Bettyblocks/feat/pie chart

ReinierVanAltena:
Basic Pie and Bar Component that can count and sum properties and display them in a chart.

Requires ApexCharts to be imported.


vdberganton:
Closing this now while working towards a solution for adding dependencies.

Also, don't directly open PR's to the acceptance branch

PR #122, opened 2020-04-29
Chart component version 0.5 basics are working

garethloot:
Chart component build on top of chart.js
It gets data by setting a model and then a label property. When you select the data component you can set colors and the data property that it needs to show. You can drag in other data sets to the top and give it there own colors and data property.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 .nvmrc
 .tool-versions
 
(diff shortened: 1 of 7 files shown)


vdberganton:
Closing this now while working towards a solution for adding dependencies

Also, don't open PRs directly to the acceptance branch

PR #133, opened 2020-05-04
Feat/hidden input

chielwester:


src/components/hiddenInput.js
```diff
@@ -0,0 +1,53 @@
...
+  allowedTypes: [],
+  orientation: 'HORIZONTAL',
+  jsx: (() => {
+    const { getActionInput, useText } = B;
+    const isDev = B.env === 'dev';
+    const { actionInputId } = options;
+    const actionInput = getActionInput(actionInputId);
+
+    return (
+      <div className={[classes.root, isDev ? classes.bordered : ''].join(' ')}>
```
vdberganton, line 14: Why always render the div in runtime only the input element is needed right

chielwester: > Why always render the div in runtime only the input element is needed right

Fixed

vdberganton:
Closing this seeing we're resolving it via a custom model

PR #140, opened 2020-05-06
Task/image option type variable ta 1497

vdberganton:


src/components/image.js
```diff
@@ -4,28 +4,30 @@
...
+
+    return (
+      <figure
+        className={[classes.figure, !imgUrl ? classes.empty : ''].join(' ')}
+      >
+        {imgUrl ? (
+          <img
+            className={classes.media}
+            src={imgUrl}
+            alt={B.Text({ value: options.imgAlt })}
```
jeroenreijs, line 18: Since this is the second time of using `B.Text`, maybe use destructuring first, like `const { Text } = B;`

src/components/appBar.js
```diff
@@ -35,12 +35,13 @@
       setAnchorEl(null);
     };
 
-    const logoComponent = logo && <img src={logo} width="100" alt="" />;
-    const logoWithLink = endpoint.id ? (
+    const logo = Text({ value: logoLink });
+    const logoCmp = logo && <img src={logo} width="100" alt="" />;
```
jeroenreijs, line 39: `logoCmp` should be written as `LogoCmp` because it is representing a React component.

src/components/cardMedia.js
```diff
@@ -4,18 +4,19 @@
...
-    const { env, useText } = B;
+    const { env, Text } = B;
     const { CardMedia } = window.MaterialUI.Core;
     const isDev = env === 'dev';
     const { type, imageSource, videoSource, iframeSource, title } = options;
 
-    const titleText = isDev
-      ? title.map(t => (t.name ? t.name : t)).join(' ')
-      : useText(title);
+    const titleText = Text(title);
```
jeroenreijs, line 12: This should be `Text({ value: title })`

jeroenreijs:
General finding:

jeroenreijs:
When you base your image on a variable, nothing shows up on the canvas. You would expect the placeholder you get when you don't use a source at all or something else that shows you used a variable.

In runtime you get this: 
image

Would be nice if you see something on the canvas as well. Now it is just a blank space.

PR #148, opened 2020-05-07
Feat/more list options

stephankaag:


jeroenreijs:
After testing it turned out the added dense option isn't added to the listitem in the list prefab. Can you please add it?

PR #149, opened 2020-05-07
Bettyblocks/feat/dialog component

bryanoemar:
@jeroenreijs I believe I have implemented your feedback. 

- Removed the wrapper div and all other unnecessary div's
- using rem unit instead of px
- using backgroundColor instead of background 

Also converted the modal into a dialog component. 

Also, I've read somewhere that the convention is to create a pull request towards the edge environment, so doing that now, instead of acceptance. 

Attempt #2 :)


src/components/dialog.js
```diff
@@ -0,0 +1,102 @@
...
+    const getSpacing = (idx, device = 'Mobile') =>
+      idx === '0' ? '0rem' : style.getSpacing(idx, device);
+    return {
+      dialog: {
+        backgroundColor: 'white',
+        margin: '0 auto',
+        padding: '1.5rem',
+        overflowWrap: 'break-word',
+      },
+      content: {
```
jeroenreijs, line 61: This class is used on the DIV to hide the Dialog in. I don't see the use of padding etc. This has no effect on the shown Dialog.
Also, the extra DIV wrapper is something we need in the page builder to add the drag/drop events, but in runtime it has no purpose. So please add it with a condition unless you are sure that you need it in runtime.

src/components/dialog.js
```diff
@@ -0,0 +1,102 @@
...
+        ].join(' ')
+      : '';
+
+    const [open, setOpen] = React.useState(false);
+    const devContent = isContentVisible && !isEmpty ? children : 'Dialog';
+
+    useEffect(() => {
+      B.defineFunction('handleDialogOpen', e => {
+        e.preventDefault();
+
```
jeroenreijs, line 29: This white line doesn't make the code better readable. Please remove it. Also, for line 35.

src/components/dialog.js
```diff
@@ -0,0 +1,102 @@
...
+        content: {
+          padding: getSpacing('M', 'Desktop'),
+        },
+      },
+      empty: {
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: () => 'center',
+        minHeight: '4rem',
+        height: '100%',
```
jeroenreijs, line 89: This height of 100% makes a column show a scrollbar. Please think of another solution. I guess the min height of 4rem will do. The empty class name is meant for showing users they dragged a container but didn't drag any children in it.

src/prefabs/dialog.js
```diff
@@ -0,0 +1,19 @@
...
+  name: 'Dialog',
+  icon: 'PanelIcon',
+  category: 'LAYOUT',
+  structure: [
+    {
+      name: 'Dialog',
+      options: [
+        {
+          value: true,
+          label: 'Visible in dev?',
```
jeroenreijs, line 11: Users probably don't understand the meaning of `dev`.

src/components/dialog.js
```diff
@@ -0,0 +1,102 @@
...
+        setOpen(false);
+      });
+    }, []);
+
+    return (
+      <div className={classNames}>
+        {isDev ? devContent : ''}
+
+        <Dialog open={open} onClose={() => setOpen(false)}>
+          <div className={classes.dialog}>{children}</div>
```
jeroenreijs, line 45: Why are you using `<div className={classes.dialog}>` to show the children?
Material UI offers `<DialogTitle>`, `<DialogContent>`, `<DialogContentText>`, `<DialogActions>` to show content in. Better use those instead.

bryanoemar:
> I think using the dialog is better than using the modal. This gives us much more options and control. However, there are no options used. Material UI offers a lot of options to configure the dialog. Why didn't you use any of those? https://material-ui.com/components/dialogs/

I believe taking things step by step. The dialog is already functional and can be worked with. In a later iteration we can extend this with other options if needed.

src/components/dialog.js
bryanoemar: >This class is used on the DIV to hide the Dialog in. I don't see the use of padding etc. This has no effect on the shown Dialog.

Thanks for pointing that out. I'll be fixing it.


> Also, the extra DIV wrapper is something we need in the page builder to add the drag/drop events, but in runtime it has no purpose. So please add it with a condition unless you are sure that you need it in runtime.

I've tested this and it gave me no errors. Everything is working fine. I don't think I need an extra div, unless I'm missing something?

src/prefabs/dialog.js
bryanoemar: Oki, I'll rename this to 'Visible' just like in the Alert component :)

src/components/dialog.js
bryanoemar: I'd like to add the options in a next iteration, since the dialog is functional as is and you are able to add title and other content just by dragging and dropping them into the dialog. It seems like a good solution. The options that Material UI offers might make it more simple for the end user, but that sounds like a nice to have for a later iteration.

vdberganton:
Closing this because we added a dialog component in a different branch!

PR #218, opened 2020-06-05
Update step.js

frankverhorst:
The wrapper class is undefined therefore obsolete.


vdberganton:
Closing because its directly going to acceptance and will be fixed in a different branch.

PR #225, opened 2020-06-09
Horizontal navbar/prefab

bryanoemar:
Option flat added, this is an option we often use at Services.


vdberganton:
Closing due to being inactive

PR #239, opened 2020-06-12
Add Change triggerEvent to Radio component

DylanCarver:


src/components/radio.js
```diff
@@ -95,6 +95,7 @@
     }
 
     const handleChange = evt => {
+      B.triggerEvent('Change', evt);
```
vdberganton, line 98: Can you change this to "onChange" so that it is consistent with the form components triggers.

After that you can also open a PR to acceptance seeing its from a forked branch

vdberganton:
Closing due to inaction

PR #253, opened 2020-06-16
Alert/prefabs

bryanoemar:
Added some options for alignment purposes which we need on projects. 
Also made a distinction between coloring of text and icon.
Furthermore, we've added a border, which can be colored depending on the status.


jeroenreijs:
These are very good suggestions. I see even more room for improvement here, so let me start a branch with these improvements and then I will add the suggested options as well.
For example the basic alignment is wrong because of an extra div that is rendered for the title.
I will let you know when it's there.

jeroenreijs:
Internal PR is this: https://github.com/bettyblocks/material-ui-component-set/pull/257
I will close your PR now.

PR #269, opened 2020-06-19
Add clear interaction to autocomplete component

stephankaag:


jeroenreijs:
You can check the current edge branch. We added a fix last week, which was merged this morning. When it works like it should, you can close this PR.

PR #302, opened 2020-06-30
Feature/nested property search and sort

thomasvanophem-betty:


src/components/dataList.js
```diff
@@ -53,104 +75,150 @@
...
+        };
+
+        const isObject = item =>
+          item && typeof item === 'object' && !Array.isArray(item);
+        const deepMerge = (target, ...sources) => {
+          if (!sources.length) return target;
+          const source = sources.shift();
+
+          if (isObject(target) && isObject(source)) {
+            // eslint-disable-next-line no-restricted-syntax
```
jeroenreijs, line 99: You can probably solve this by using Object.entries to avoid using disabling the linter.
I didn't try this one yet. For the others to come I will include examples.

src/components/dataList.js
```diff
@@ -53,104 +75,150 @@
...
+
+        if (!isDev && orderBy) {
+          orderByArray.reduce((acc, property, index) => {
+            const prop = getProperty(property);
+            // eslint-disable-next-line no-return-assign
+            return (acc[prop.name] =
+              index !== orderByArray.length - 1 ? {} : order.toUpperCase());
+          }, sort);
+        }
+
```
jeroenreijs, line 124: This part, can be written differently using reduceRight.

```
        const sort =
          !isDev && orderBy
            ? orderByArray.reduceRight((acc, property, index) => {
                const prop = getProperty(property);
                return index === orderByArray.length - 1
                  ? { [prop.name]: order.toUpperCase() }
                  : { [prop.name]: acc };
              }, {})
            : {};

```
And you can do this in datatable as well.

src/components/dataList.js
```diff
@@ -53,104 +75,150 @@
...
+            searchPropertyArray.reduce(
+              // eslint-disable-next-line no-return-assign
+              (acc, property, index) =>
+                (acc[property] =
+                  index !== searchPropertyArray.length - 1
+                    ? {}
+                    : { matches: search }),
+              searchFilter,
+            );
+          }
```
jeroenreijs, line 142: This part can become:

```
          const searchFilter = searchProperty
            ? searchPropertyArray.reduceRight(
                (acc, property, index) =>
                  index === searchPropertyArray.length - 1
                    ? { [property]: { matches: search } }
                    : { [property]: acc },
                {},
              )
            : {};

```
And you can do this in datatable as well.

src/components/dataTableColumn.js
```diff
@@ -48,6 +47,17 @@
       handleSort(prop, sortOrder);
     };
 
+    const isFilterSelected = fields => {
+      if (!fields || fields.length !== propertyArray.length) return false;
+
+      // eslint-disable-next-line no-plusplus
+      for (let index = 0; index < fields.length; index++) {
```
jeroenreijs, line 54: Instead of `index++` you can use `index += 1` to avoid disabling the linter.

src/components/dataList.js
thomasvanophem-betty: Fixed this one with Object.keys instead of Object.entries 
`Object.entries({ "name": "Jeroen" })[0]` results in `["name", "Jeroen"]` which in turn results in a key `"Jeroen"` in the returned object. `Object.keys` only returns the actual keys of the source object.

PR #303, opened 2020-06-30
Feature/submit button spinner

thomasvanophem-betty:


jeroenreijs:
Please never merge the edge branch into yours. This could mean that you have something in your branch that we don't want in acceptance (yet). See this commit: https://github.com/bettyblocks/material-ui-component-set/pull/303/commits/981b2ae998d0f929d711e5ee5d1f9649fa0e3377

thomasvanophem-betty:
Hmm.. didn't think that one completely through.. I saw a merge conflict and thought.. let's fix it 😅  Will fix and won't happen again.

src/components/button.js
```diff
@@ -22,8 +22,8 @@
       actionId,
       buttonText,
     } = options;
-
-    const { env, useText } = B;
+    const { loading: parentLoading } = parent || {};
```
vdberganton, line 25: Why do this with prop drilling if you can achieve this with a triggerEvent?

thomasvanophem-betty: Because the form already passes the loading state to it's children. If you prefer a triggerEvent I can change that.

vdberganton: Yes please, that is probably before we had triggerEvent and needs to be changed.

PR #310, opened 2020-07-07
fix: order by relation in datatable

jeroenreijs:


jeroenreijs:
Closing this. This is the same solution as https://github.com/bettyblocks/material-ui-component-set/pull/302, but this latter one has more changes.

PR #349, opened 2020-07-17
feat: change the button and list to use hooks

wmahad:


src/components/button.js
```diff
@@ -77,55 +77,37 @@
...
+    const [actionCallback, { loading }] = useAction(actionId);
+    if (isAction) onClickFn = actionCallback;
 
-    if (isAction) {
-      ButtonComponent = (
-        <B.Action actionId={actionId}>
-          {(callAction, { loading }) => {
-            const onClickAction = event => {
-              event.preventDefault();
-              if (!isDev && !loading && linkType === 'action') {
```
vdberganton, line 120: Is this check not missing now? So now doing multiple clicks in a row might give unexpected behaviour?

wmahad: IMO, it was unnecessary. isAction in itself is sufficient enough.

PR #360, opened 2020-07-23
fix preset dat value to make sure it works out of the box

garethloot:
The default values in the date time picker fixed in a format that works out of the box


vdberganton:
We're going to solve this by sending a ISO value and this was form a forked branch directly in acceptance so closing this branch.

PR #416, opened 2020-09-09
Fix/enable to select variable in app bar

Angelodaniel:
Added useText to appbar, so it can receive variables instead of breaking


jeroenreijs:
Please change the target branch to `edge`.

Angelodaniel:
@jeroenreijs Changed it do Edge

jeroenreijs (approved):

PR #438, opened 2020-09-17
Task/introduce auth profile option top 1227

M-Ryan92:


src/components/dataTable.js
```diff
@@ -58,7 +60,7 @@
     const [searchTerm, setSearchTerm] = useState('');
     const searchPropertyArray = [searchProperty].flat();
```
RobinBetty, line 61: Do we still use this after the removal in the getProperty

jeroenreijs (approved):

PR #439, opened 2020-09-17
Feat/prevent decimal if number

DylanCarver:
Prevent decimals from being typed into the number input field.


DylanCarver:
This fix has been made in another branch.

PR #443, opened 2020-09-21
fix: priceInput value type

DylanCarver:
Change `priceInput` type from number to decimal.


DylanCarver:

PR #444, opened 2020-09-21
fix: timePicker icon fix

DylanCarver:
When the type (defined in the prefab) is `time` an `<AccessTime />` icon will be shown instead of the `<Event />` icon. 
Previously:
Screen Shot 2020-09-21 at 18 45 39
Now:
Screen Shot 2020-09-21 at 18 45 59


DylanCarver:

PR #445, opened 2020-09-21
fix: add use24HourClockTime key back to timePicker

DylanCarver:
Add optional 24 hour format back to timePicker component.


DylanCarver:

PR #453, opened 2020-09-24
feat: allow linking between data container and list and table views

benjamin-rocks:
- add logic to data container to read from currentRecord
   only reads from option if no filter was manually set
- add rowClick event to data list
- add currentRow option to data list
- add currentRow option to data table
- add new default global function

### DIFF
--- a/src/components/dataContainer.js
+++ b/src/components/dataContainer.js
@@ -53,10 +53,8 @@
             return builderLayout();
           }
 
-          const dynamicFilter = getFilter();
-
           return (
-            <B.GetOne modelId={model} filter={dynamicFilter}>
+            <B.GetOne modelId={model} filter={getFilter()}>
               {({ loading, error, data }) => {
                 if (loading) {
                   B.triggerEvent('onLoad', loading);
--- a/src/components/dataContainer.js
+++ b/src/components/dataContainer.js
@@ -15,8 +15,20 @@
           authProfile,
           redirectWithoutResult,
           showError,
+          currentRecord,
         } = options;
         const displayError = showError === 'built-in';
+
+        const getFilter = React.useCallback(() => {
+          if (isDev || !currentRecord || !model) {
+            return filter;
+          }
+
+          const idProperty = B.getIdProperty(model);
+          return {
+            [idProperty.id]: { eq: currentRecord },
+          };
+        }, [isDev, filter, currentRecord, model]);
 
         const builderLayout = () => (
           <>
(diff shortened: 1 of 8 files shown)


dehlak (dismissed):

src/components/dataContainer.js
```diff
@@ -41,8 +53,11 @@
             return builderLayout();
           }
 
+          const dynamicFilter = getFilter();
+          console.log({ dynamicFilter, currentRecord });
```
NiekBetty, line 57: Wut 😜

NiekBetty:
How can I test the `currentRecord` option in the `DataContainer`?

src/components/dataContainer.js
```diff
@@ -15,9 +15,21 @@
           authProfile,
           redirectWithoutResult,
           showError,
+          currentRecord,
```
NiekBetty, line 18: How can I test the `currentRecord` option? Do you have a steps to reproduce? :)

src/components/dataList.js
```diff
@@ -118,10 +118,24 @@
           repeat();
         });
 
+        React.useEffect(() => {
+          if (isDev) return;
+          if (options.currentRecord !== '') {
+            B.triggerEvent('OnRowClick', options.currentRecord);
```
NiekBetty, line 124: Do we need a `useEffect` for this? Is it possible to add this to the `handleClick` function?

benjamin-rocks: We should always use a `useEffect` because setingState is not async. Before hooks this was acheived by passing a second argument to `setState(mutation, callback)`. But with `useState()` useEffects are the best way to perform actions "after" state has changed.

src/components/dataContainer.js
benjamin-rocks: Set the new global interaction that we have added. It looks like this:

You have a DataTable and a DataContainer. **BOTH** have should have their model option configured.

Next set up this interaction

[DataTable] OnRowClick
[InteractionType] Global
[Interaction] SetCurrentRecord
[Source] DataTable.currentOption
[Target] DataContainer.currentOption

press `Play` now you should see that the DataContainer's currentRecord is changing to match the row that was clicked!

NiekBetty: Ah cool!

NiekBetty (dismissed):

jeroenreijs (dismissed):

jeroenreijs (approved):

PR #460, opened 2020-09-30
Add allow overflow option to Column

garethloot:
When you nest items in a column that use shadows or relative positioning the column wil cut these of because of overflow settings. This adds a Toggle to allow overflow on the column


jeroenreijs:
There are more prefabs which use the column: Dialog, Panel and Paper. Please fix those as well at least for consistency, and in same cases it might break.

garethloot:
Ah yeah forgot those. There fixed now.

jeroenreijs (approved):

PR #474, opened 2020-10-05
Add onActionDone interaction to form component; Removed 100px fixed height in dialog; Changed submit button default color to primary

MarcelKorporaal:
onActionDone was already PR last week. Added some tiny changes to 2 prefabs as well.

### DIFF
--- a/src/components/form.js
+++ b/src/components/form.js
@@ -108,6 +108,9 @@
         };
 
         const trigger = (data, loading, error) => {
+          if (data || error) {
+            B.triggerEvent('onActionDone');
+          }
           if (data) {
             B.triggerEvent('onActionSuccess', data.actionb5);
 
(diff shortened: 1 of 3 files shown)


src/components/form.js
```diff
@@ -108,6 +108,9 @@
         };
 
         const trigger = (data, loading, error) => {
+          if (data || error) {
+            B.triggerEvent('onActionDone');
```
jeroenreijs, line 112: We have `onActionSuccess` when there is data and `onActionError` when there is an error. Why should we need this extra event?

jeroenreijs (changes requested):
Please use `edge` as target branch.

src/components/form.js
MarcelKorporaal: It will save a lot of extra interactions. For instance, wether there's an error or success doesn't matter: you always wanna remove the loading state of the button when the submit is done.

PR #475, opened 2020-10-05
Added 3 new components Badge, Link and Snackbar

Reinier1982:
3 New Components Badge, Link and Snackbar
All official Material UI components 

Badge for showing the number of items on an icon or button.
Link for making an icon or any other component linkable.
Snackbar for making alerts and other messages show in a small pop-up (snackbar).

### DIFF
--- a/src/components/badge.js
+++ b/src/components/badge.js
@@ -1,0 +1,80 @@
+(() => ({
+  name: 'Badge',
+  type: 'CONTENT_COMPONENT',
+  allowedTypes: ['CONTENT_COMPONENT'],
+  orientation: 'VERTICAL',
+  jsx: (() => {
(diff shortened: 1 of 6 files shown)


chrisobdam:
Why do you have a link component?

chrisobdam:
Maybe better to create a PR per component?

Reinier1982:
> Why do you have a link component?

The Link component is so that you can let the Icon Component become a clickable component. But also so other components like images can be come clickable.

nulian:
Please don't skip the commit hooks
See: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

We use it to generate changelogs at least one commit with feat: added component
would be nice to have.

PR #481, opened 2020-10-06
Feat/submit on change of inputs

garethloot:
Add a submit on change option to Select, Autocomplete, Checkbox, Checkbox group and Radio components.
Add a submit after blur option to the textField component and to all the prefabs that use textField.

If you enable the on change or on blur option the parent form wil be triggered to submit. If there are validation errors the submit wil be canceled.

### DIFF
--- a/src/components/autoComplete.js
+++ b/src/components/autoComplete.js
@@ -28,6 +28,7 @@
       customModelAttribute: customModelAttributeObj,
       property,
       nameAttribute,
+      submit,
     } = options;
     const { Autocomplete } = window.MaterialUI.Lab;
     const {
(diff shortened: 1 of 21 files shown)


nulian:
Please don't skip the git hooks see  https://github.com/conventional-changelog/commitlint/#what-is-commitlint
We use it to generate changelogs.

garethloot:
@nulian Ah yes now i know how to create the correct commit messages. I was just skipping those pesky husky errors :)

jeroenreijs:
@garethloot Although we like this creativity, we would like to see that interactions are used for this kind of behavior. That will mean somewhat more work when someone is creating a form, but it will be more consistent with our current way of working.

PR #484, opened 2020-10-07
Feat: icon has the ability to link to pages now

Reinier1982:
Feat: icon has the ability to link to pages now


jeroenreijs:
Please create a PR for each component separately so they don't interfere with each other. Also, we're working on a snackbar component ourselves. Thanks.

PR #491, opened 2020-10-10
feat: add option to get image directly from image property

garethloot:
- Add Image sourcetype button group
- Add property id and optional size name
- Add fallback url if no image


nulian:
This is not something we could merge like this at the moment.
Because fields like propertyId and fileType as text inputs are a bit complicated.

Our preferred solution would be able to select the size either through the property selector or from a list after you select your property.

PR #524, opened 2020-10-20
Feat/add badge component new

Reinier1982:
Feat/add badge component new


vdberganton:
As discussed: 
Prefer that the badge is going to be added as an extra option on the button & icon component. 

Instead of it now functioning as a container component.

PR #531, opened 2020-10-21
feat/overflow option column

garethloot:
- add allow overflow toggle to columns to be able to overflow shadows and relative positioned children


jeroenreijs (dismissed):

vdberganton:
Can we just rename this key to `overflow` and the label to "Allow overflow". 

Instead of "Allow allowOverflow"

jeroenreijs:
Looks like there is no response here on our requests?
Target branch should be `edge`. Also this looks like a change you already added?  https://github.com/bettyblocks/material-ui-component-set/commit/f1188f6b3b15f1a484f215ea5a13cb9129fe6089 
This branch is lost in the new Travis, because of aging.
I'm closing this PR because of all these reasons.

PR #535, opened 2020-10-23
Fix/set default progress to zero

chielwester:
The progress component defaults to 50 which causes it to mistakenly display 50% when the actual progress is zero or non existent.


jeroenreijs:
Good fix, but maybe only do this in runtime, because it is nice to see something in the pagebuilder.
Please use `edge` for target branch.
Your pipeline failed because of your commit message:
https://github.com/conventional-changelog/commitlint/#what-is-commitlint

PR #541, opened 2020-10-28
feat: add overflow to box

DylanCarver:
[Related Jira ticket](https://bettyblocks.atlassian.net/jira/software/projects/TA/boards/73?selectedIssue=TA-1946).
This pr allows you to set the overflow value of a box component.
Screenshot 2020-10-28 at 11 25 07
## Reason
Another way for the streched boxes child (within a grid) to use its parents actual height, besides giving the box a height itself (of 1px/0px), is to give it an overflow value of 'hidden' (explained in more detail [here](https://stackoverflow.com/a/36247448)).

### DIFF
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -102,6 +102,7 @@
         bottom: ({ options: { bottom } }) => bottom,
         left: ({ options: { left } }) => left,
         width: ({ options: { width } }) => width,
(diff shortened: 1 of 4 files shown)


DylanCarver:
Closed as the issue is fixed in pr #544.

PR #563, opened 2020-10-30
Feat/add auto submit option to form

chielwester:
Add a trigger on component load to form and interaction to submit the form. This way you can automatically submit the form on component load. Used this for single sign on interactions in oauth flows.

### DIFF
--- a/src/components/form.js
+++ b/src/components/form.js
@@ -6,7 +6,14 @@
   jsx: (
     <div>
       {(() => {
-        const { env, Children, Action, useAllQuery, getActionInput } = B;
+        const {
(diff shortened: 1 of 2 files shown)


jeroenreijs:
It is no longer needed to put `B.defineFunction` in a `useEffect`. Also, can you rename the event `onComponentLoaded` to `onComponentRendered`.

chielwester:
I've changed the trigger name and put the interaction outside of the useEffect. The trigger needs to remain inside a useEffect otherwise this results into an error

jeroenreijs (approved):

PR #617, opened 2020-11-11
fix: data container shows first record in case of no filter

simonedebest-bb:


src/components/dataContainer.js
```diff
@@ -56,7 +57,7 @@
         }
 
         const CanvasLayout = () => {
-          if (!model) {
+          if (!model || !hasFilter) {
```
nulian, line 60: You can just check !hasFilter here.

jeroenreijs (approved):

PR #633, opened 2020-11-16
feat: add external link option to listitem

JorisPannekeet:


jeroenreijs:
Seeing this build failed and the fact that we just added the external link to the listitem (currently on edge) you might just want to reconsider and close this PR.

PR #636, opened 2020-11-17
feat: add logo width to appbar

JorisPannekeet:


jeroenreijs:
Width with unit other than `px` doesn't work like this. So you'll have to add a className in which you cover the width option.

PR #670, opened 2020-11-30
feat: button has the option for a tooltip

Reinier1982:


jeroenreijs:
Cool feature! Can you show the tooltip on the canvas as well, maybe by adding another toggle option? Because now you have to view the page in runtime to see the tooltip.
And it would be nice if you can change the color of the tooltip.

PR #684, opened 2020-12-03
Feat/translate data table search label

chielwester:


src/components/dataTable.js
```diff
@@ -559,6 +562,9 @@
               classes={{ root: classes.pagination }}
               rowsPerPageOptions={[5, 10, 25, 50, 100]}
               labelRowsPerPage={useText(labelRowsPerPage)}
+              labelDisplayedRows={({ from, to, count }) =>
+                `${from}-${to} van ${count !== -1 ? count : `meer dan ${to}`}`
```
jeroenreijs, line 566: You've made the search field placeholder configurable, but here I see some hardcoded dutch words. That is not what you had in mind I assume?

chielwester: You are right about this I was a bit too fast here. Changed it in a new commit

PR #685, opened 2020-12-07
feat: add table row

benjamin-rocks:
you may now add interactions directly to table rows.

this also means you have access to properties just as you do with any other component 🎉


benjamin-rocks:
This feature has been superseded by a better solution

PR #691, opened 2020-12-10
fix: handle rowclick with outside variables

jeroenreijs:


jeroenreijs:
These changes interfere with changes to come, so for now I will close this ticket.

PR #763, opened 2021-01-05
feat(techsub-3022): conditional table columns

garethloot:
- add isFalse and isTrue triggers to conditional component
- add visibilty interactions to tableColumn

### DIFF
--- a/src/components/conditional.js
+++ b/src/components/conditional.js
@@ -55,7 +55,7 @@
 
         B.defineFunction('Hide', () => setVisible(false));
         B.defineFunction('Show', () => setVisible(true));
-        B.defineFunction('ToggleVisibility', () => setVisible(s => !s));
+        B.defineFunction('Show/Hide', () => setVisible(s => !s));
 
         if (!isDev && !visible) return null;
         return isPristine ? 'Conditional' : children;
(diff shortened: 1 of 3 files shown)


src/components/conditional.js
```diff
@@ -45,9 +45,17 @@
...
+            B.triggerEvent('isTrue', true);
+          } else {
+            B.triggerEvent('isFalse', false);
+          }
+        }, [visible]);
+
         B.defineFunction('Hide', () => setVisible(false));
         B.defineFunction('Show', () => setVisible(true));
-        B.defineFunction('Show/Hide', () => setVisible(s => !s));
+        B.defineFunction('ToggleVisibility', () => setVisible(s => !s));
```
vdberganton, line 58: Please don't rename the interactions without consulting UX.

src/components/conditional.js
```diff
@@ -45,9 +45,17 @@
           setVisible(initialVisibility);
         }, [checkCondition]);
 
+        useEffect(() => {
+          if (visible) {
+            B.triggerEvent('isTrue', true);
+          } else {
+            B.triggerEvent('isFalse', false);
+          }
+        }, [visible]);
```
vdberganton, line 54: This is very nice to be able to build an if else with multiple conditionals, kudos!

src/components/conditional.js
garethloot: Ow that means i was maybe a commit behind when i created the branch. Dit not do that on purpose.

jeroenreijs (approved):

PR #772, opened 2021-01-07
Feat/preview uploaded image work 42

DylanCarver:
Using JSS's `extend` doesn't seem to work when using dynamic options from the prefab.. for example:
```jsx 
class1: {
  width: ({ options: { imagePreviewWidth } }) => imagePreviewWidth,
},
class2: {
  extend: 'image',
  // the width will not be given through to this class2
}
```
This is why the width and height properties are set quite a few times in the style.

[Related Jira ticket](https://bettyblocks.atlassian.net/jira/software/projects/WORK/boards/471?selectedIssue=WORK-42)

# Testing
## Using image previews in the component itself
- Drag the ImageUpload prefab to the canvas
- There should be some new options to configure (type, show image preview & image preview width/height)
Screenshot 2021-01-14 at 09 43 08
- These options should affect the dev-env preview as well as the run-time after compiling.

## Using image previews via interactions
When the 'allow multiple' checkbox is turned off, you're able to preview the image in a separate box component (as a background image) or media component using interactions.
- drag an imageUpload, box and media component to the canvas.
- uncheck 'show image preview' in the imageUpload, as we will be displaying the image in the other components.
- add 2 new interactions via the imageUpload prefab. 
- for the first interaction, use these options to display the image on a box component:
Screenshot 2021-01-14 at 09 51 16
- for the second interaction use these options to show the image on a media component
Screenshot 2021-01-14 at 09 52 08
- compile the page and upload an image, the image should be displayed on the box and media components.

### DIFF
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -25,26 +25,10 @@
     const opac = transparent ? 0 : 1;
     const [opacity, setOpacity] = useState(opac);
     const [interactionBackground, setInteractionBackground] = useState('');
-    const [boxStyles, setBoxStyles] = useState({});
 
     B.defineFunction('setBackgroundImage', url => {
       setInteractionBackground(url);
     });
-
-    useEffect(() => {
-      setBoxStyles({ opacity });
-    }, []);
-
-    useEffect(() => {
-      if (interactionBackground) {
-        setBoxStyles({
-          backgroundImage: interactionBackground
-            ? `url("${interactionBackground}")`
-            : 'none',
-          opacity,
-        });
-      }
-    }, [interactionBackground]);
 
     const boxOptions = {
       display: isFlex && 'flex',
@@ -78,7 +62,12 @@
         onClick={handleClick}
         onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
-        style={boxStyles}
+        style={{
+          backgroundImage: interactionBackground
+            ? `url("${interactionBackground}")`
+            : 'auto',
+          opacity,
+        }}
       >
         {isEmpty ? 'Box' : children}
       </Box>
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -24,6 +24,27 @@
     const isFlex = alignment !== 'none' || valignment !== 'none';
     const opac = transparent ? 0 : 1;
     const [opacity, setOpacity] = useState(opac);
+    const [interactionBackground, setInteractionBackground] = useState('');
+    const [boxStyles, setBoxStyles] = useState({});
+
+    B.defineFunction('setBackgroundImage', url => {
+      setInteractionBackground(url);
+    });
+
+    useEffect(() => {
+      setBoxStyles({ opacity });
+    }, []);
+
+    useEffect(() => {
+      if (interactionBackground) {
+        setBoxStyles({
+          backgroundImage: interactionBackground
+            ? `url("${interactionBackground}")`
+            : 'none',
+          opacity,
+        });
+      }
+    }, [interactionBackground]);
 
     const boxOptions = {
       display: isFlex && 'flex',
@@ -57,7 +78,7 @@
         onClick={handleClick}
         onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
-        style={{ opacity }}
+        style={boxStyles}
       >
         {isEmpty ? 'Box' : children}
       </Box>
--- a/src/components/fileUpload.js
+++ b/src/components/fileUpload.js
@@ -199,7 +199,11 @@
         <IconButton
           size="small"
           className={classes.remove}
-          onClick={file ? () => removeFileFromList(file.url) : {}}
+          onClick={() => {
+            if (file) {
+              removeFileFromList(file.url);
+            }
+          }}
         >
           <Delete className={classes.deleteIcon} fontSize="small" />
         </IconButton>
@@ -226,17 +230,15 @@
       switch (type) {
         case 'grid':
           return (
-            <>
-              <div className={classes.gridView}>
-                <div className={classes.gridItem}>
-                  {showImagePreview && <div className={classes.gridDevImage} />}
-                  <div className={classes.gridItemDetails}>
-                    <FileDetails />
-                    <DeleteButton />
-                  </div>
+            <div className={classes.gridView}>
+              <div className={classes.gridItem}>
+                {showImagePreview && <div className={classes.gridDevImage} />}
+                <div className={classes.gridItemDetails}>
+                  <FileDetails />
+                  <DeleteButton />
                 </div>
               </div>
-            </>
+            </div>
           );
         case 'list':
         default:
@@ -267,28 +269,26 @@
       switch (type) {
         case 'grid':
           return (
-            <>
-              <div className={classes.gridView}>
-                <div className={classes.gridItem}>
-                  {showImagePreview && (
-                    <div
-                      style={{
-                        backgroundImage: `url("${file.url}")`,
-                      }}
-                      className={classes.gridImage}
-                    />
-                  )}
-                  <div className={classes.gridItemDetails}>
-                    <FileDetails
-                      file={file}
-                      fileType={uploadedFile.type}
-                      fileSize={uploadedFile.size}
-                    />
-                    <DeleteButton file={file} />
-                  </div>
+            <div className={classes.gridView}>
+              <div className={classes.gridItem}>
+                {showImagePreview && (
+                  <div
+                    style={{
+                      backgroundImage: `url("${file.url}")`,
+                    }}
+                    className={classes.gridImage}
+                  />
+                )}
+                <div className={classes.gridItemDetails}>
+                  <FileDetails
+                    file={file}
+                    fileType={uploadedFile.type}
+                    fileSize={uploadedFile.size}
+                  />
+                  <DeleteButton file={file} />
                 </div>
               </div>
-            </>
+            </div>
           );
         case 'list':
         default:
@@ -322,20 +322,18 @@
       switch (type) {
         case 'grid':
           return (
-            <>
-              <div className={classes.gridView}>
-                <div className={classes.gridItem}>
-                  {showImagePreview && (
-                    <div className={classes.gridUploadingImage}>
-                      <CloudUpload />
-                    </div>
-                  )}
-                  <div className={classes.gridItemDetails}>
-                    <span>Uploading</span>
+            <div className={classes.gridView}>
+              <div className={classes.gridItem}>
+                {showImagePreview && (
+                  <div className={classes.gridUploadingImage}>
+                    <CloudUpload />
                   </div>
+                )}
+                <div className={classes.gridItemDetails}>
+                  <span>Uploading</span>
                 </div>
               </div>
-            </>
+            </div>
           );
         case 'list':
         default:
@@ -357,7 +355,7 @@
       }
     };
 
-    const Control = (
+    const Control = () => (
       <FormControl
         fullWidth={fullWidth}
         required={required}
@@ -376,7 +374,6 @@
         <FormHelperText classes={{ root: classes.helper }}>
           {helperValue}
         </FormHelperText>
-        {loading && B.triggerEvent('onLoad')}
         <div className={classes.messageContainer}>
           {data &&
             data.length > 0 &&
@@ -385,6 +382,12 @@
         </div>
       </FormControl>
     );
+
+    useEffect(() => {
+      if (loading) {
+        B.triggerEvent('onLoad');
+      }
+    }, [loading]);
 
     useEffect(() => {
       if (files.length > 0) {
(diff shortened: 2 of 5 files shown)


src/components/box.js
```diff
@@ -24,6 +24,27 @@
...
+      setBoxStyles({ opacity });
+    }, []);
+
+    useEffect(() => {
+      if (interactionBackground) {
+        setBoxStyles({
+          backgroundImage: interactionBackground
+            ? `url("${interactionBackground}")`
+            : 'none',
+          opacity,
```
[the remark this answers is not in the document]
NiekBetty: Also FYI, setting state within a `useEffect` should be treated very carefully and by default the linter warns you about it, check: `eslint react-hooks/exhaustive-deps`

src/components/fileUpload.js
```diff
@@ -180,6 +192,171 @@
...
     );
 
+    const Hr = () => <hr className={classes.hr} />;
+
+    const DeleteButton = ({ file }) => (
+      <div className={classes.deleteButtonWrapper}>
+        <IconButton
+          size="small"
+          className={classes.remove}
+          onClick={file ? () => removeFileFromList(file.url) : {}}
```
NiekBetty, line 202: Are you sure `onClick` can receive an object as value? I think the if statement should be moved inside the function like this:

```javascrip
onClick={() => {
  if(file) {
    removeFileFromList(file.url);
  }
}}
```

src/components/fileUpload.js
```diff
@@ -180,6 +192,171 @@
...
+      );
+
+      if (!multiple) {
+        B.triggerEvent('onFileUpload', file.url);
+      }
+
+      switch (type) {
+        case 'grid':
+          return (
+            <>
```
NiekBetty, line 270: This fragment isn't needed.

src/components/fileUpload.js
```diff
@@ -180,6 +192,171 @@
...
+            </>
+          );
+      }
+    };
+
+    const UploadingFile = () => {
+      switch (type) {
+        case 'grid':
+          return (
+            <>
```
NiekBetty, line 325: This fragment isn't needed.

src/components/fileUpload.js
```diff
@@ -200,24 +377,12 @@
           {helperValue}
         </FormHelperText>
         {loading && B.triggerEvent('onLoad')}
```
NiekBetty, line 379: I know this is not your code, but this should be done in a `useEffect` hook. This is no JSX 😁 . Can you move this? :)

src/components/fileUpload.js
```diff
@@ -229,7 +394,14 @@
 
     B.defineFunction('clearFileUpload', e => clearFiles(e));
 
-    return isDev ? <div className={classes.root}>{Control}</div> : Control;
+    return isDev ? (
+      <div>
```
NiekBetty, line 398: This can probably be a fragment?

src/components/fileUpload.js
```diff
@@ -229,7 +394,14 @@
...
+        <div className={classes.root}>{Control}</div>
+        <DevUploadedFile />
+      </div>
+    ) : (
+      Control
```
NiekBetty, line 403: This looks a bit weird to me. I think it should be: `<Control />`?

src/components/fileUpload.js
```diff
@@ -285,6 +457,7 @@
       span: {
         flex: 1,
         textAlign: 'start',
+        marginBottom: '0.1875rem!important',
```
NiekBetty, line 460: On the next line the same kind of styling is applied, but it is written in an array. Can you choose either one an make them consistent? :)

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'space-between',
+      },
+      gridView: {
+        display: 'flex',
+      },
+      gridItem: {
+        display: 'flex',
+        borderRadius: '0.3rem',
```
NiekBetty, line 511: Rem values should always end with a `5`. In this case `0,3 * 16 = 4,8`. I don't think you want to render 4,8px? 🤔 Or do I miss something here?

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        flexDirection: 'column',
+        border: '1px solid #eee',
```
NiekBetty, line 513: Can you convert this to rem?

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        marginRight: '0.9375rem',
+        marginBottom: '0.9375rem',
```
NiekBetty, line 515: Might as well make it `1rem` I guess? As I know Thomas always works with spacing of 2, 4, 8, 12 or 16px.

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+      },
+      gridItemDetails: {
+        maxWidth: ({ options: { imagePreviewWidth, showImagePreview } }) =>
+          showImagePreview ? imagePreviewWidth : 'auto',
+        display: 'flex',
+        margin: '0.9375rem',
```
NiekBetty, line 521: Same here I guess, I would use `1rem`

NiekBetty: Can you check the other places where `0.9375rem` is used as well? :)

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        justifyContent: 'space-between',
+      },
+      devImage: {
+        display: 'flex',
+        overflow: 'hidden',
+        margin: '0.9375rem',
+        borderStyle: 'dashed',
+        borderColor: '#AFB5C8',
+        borderRadius: '0.3rem',
```
NiekBetty, line 530: Rem values should end with the number 5

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        borderWidth: '0.0625rem',
+        border: '2px dashed black',
```
NiekBetty, line 532: Can you convert this to rem?

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        backgroundColor: '#F0F1F5',
+        width: ({ options: { imagePreviewWidth } }) => imagePreviewWidth,
+        height: ({ options: { imagePreviewHeight } }) => imagePreviewHeight,
+      },
+      gridDevImage: {
+        extend: 'devImage',
+
+        margin: 0,
+        borderRadius: '0.3rem 0.3rem 0 0',
```
NiekBetty, line 541: Rem values should end with 5

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        width: ({ options: { imagePreviewWidth } }) => imagePreviewWidth,
+        height: ({ options: { imagePreviewHeight } }) => imagePreviewHeight,
+      },
+      deleteIcon: {
+        color: `${t.colors.light}!important`,
```
NiekBetty, line 546: Can you apply the same consistency here as I asked in a previous comment with the margins? Either string interpolate or use the array notation on all of them.

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+      },
+      uploadingImage: {
+        border: 'none',
+        color: 'white',
+        display: 'flex',
+        overflow: 'hidden',
+        margin: '0.9375rem',
+        alignItems: 'center',
+        borderRadius: '0.3rem',
```
NiekBetty, line 555: Rem values should en with a 5. Can you check all rem values?

src/components/fileUpload.js
```diff
@@ -322,12 +498,129 @@
...
+        width: ({ options: { imagePreviewWidth } }) => imagePreviewWidth,
+        height: ({ options: { imagePreviewHeight } }) => imagePreviewHeight,
+      },
+      fileDetails: {
+        flexGrow: 1,
+        maxWidth: ({
+          options: { type, imagePreviewWidth, showImagePreview },
+        }) =>
+          showImagePreview && type === 'grid'
+            ? `calc(${imagePreviewWidth} - 58px)`
```
NiekBetty, line 591: Question: Do you need the `calc`? What type is `imagePreviewWidth`? A string or a number?

src/components/media.js
```diff
@@ -16,14 +16,32 @@
     } = options;
 
     const titleText = useText(title);
-    const imgUrl = useText(imageSource);
     const videoUrl = useText(videoSource);
     const iframeUrl = useText(iframeSource);
+    const [imgUrl, setImgUrl] = useState(useText(imageSource));
```
NiekBetty, line 21: I don't think you're allowed to use a hook this way (https://reactjs.org/docs/hooks-rules.html). Can you move the `useText` outside of the `useState`, assign the value to a variable and pass that variable to `useState`?

src/components/media.js
```diff
@@ -16,14 +16,32 @@
...
+
+    useEffect(() => {
+      setImgUrl(useText(imageSource));
+    }, [imageSource]);
+
+    useEffect(() => {
+      setIsImage(type === 'img' && imgUrl);
+    }, [type, imgUrl]);
+
+    useEffect(() => {
```
NiekBetty, line 36: Are all these useEffects really necessary?

NiekBetty: Can you explain why they are?

NiekBetty (changes requested):

NiekBetty:
I found a bug:

**Steps to reproduce:**
1. Add imageUpload component to canvas
2. Go to runtime
3. Select a file
4. Remove the file by clicking on the trash can
5. Select a new file

**Expected result:**
Image preview of the second image should be visible

**Actual result:**
No image preview

NiekBetty:
Bug:

**Steps to reproduce:**
1. Setup the image upload in the Box and media component according to the description in this PR.
2. Go to the runtime
3. Select an image
4. Remove the selected image

**Expected result:**
The previews should not display the image anymore

**Actual result:**
The previews still show the image.

src/components/fileUpload.js
DylanCarver: This one must be a div so that the component is selectable within the builder.

src/components/fileUpload.js
DylanCarver: In order to use `<Control />` in that way, I've changed the way that `Control` was defined, from `const Control = (...)` to `const Control = () => (...)`.

src/components/fileUpload.js
DylanCarver: I'll change them to string interpolation, as the array notation wasn't working as done on the next line.

src/components/fileUpload.js
DylanCarver: Good one. It's a string, but I'm able to parse it via js: `${parseInt(imagePreviewWidth, 10) - 58}px`, which I believe would be nicer.

src/components/fileUpload.js
DylanCarver: You're correct. I miscalculated here. I'll fix the rem values for the rest of the doc. 👍🏻

src/components/box.js
```diff
@@ -24,6 +24,27 @@
     const isFlex = alignment !== 'none' || valignment !== 'none';
     const opac = transparent ? 0 : 1;
     const [opacity, setOpacity] = useState(opac);
+    const [interactionBackground, setInteractionBackground] = useState('');
+    const [boxStyles, setBoxStyles] = useState({});
```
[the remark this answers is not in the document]
DylanCarver: Turns out it wasn't necessary indeed. It was initially because I wanted to prevent from having the `backgroundImage` from being set to `none` when no image was set via the interaction, but setting it directly as the style option and changing the `none` to `auto` fixed this issue.

src/components/box.js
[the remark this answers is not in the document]
DylanCarver: Thanks for letting me know! I'll definitely read up on this. although my linter didn't notify me about this one 🤔

DylanCarver:
> I found a bug:
> 
> **Steps to reproduce:**
> 
> 1. Add imageUpload component to canvas
> 2. Go to runtime
> 3. Select a file
> 4. Remove the file by clicking on the trash can
> 5. Select a new file
> 
> **Expected result:**
> Image preview of the second image should be visible
> 
> **Actual result:**
> No image preview

As mentioned on Slack, another ticket will be created regarding this issue, as it is not a result of these changes, and already existed before on the fileUpload component.

src/components/fileUpload.js
NiekBetty: Ah got it 👌

src/components/fileUpload.js
NiekBetty: Hmm if it is a string I guess the unit is in there as well? `${parseInt(imagePreviewWidth, 10) - 58}px` won't work if the `imagePreviewWidth` is for example `1rem`. Your original solution will work.

DylanCarver: Good point. I've reverted it back to `calc`.

NiekBetty (approved):

PR #787, opened 2021-01-13
feat: add switch component

chrisobdam:


src/prefabs/switch.js
```diff
@@ -0,0 +1,194 @@
+(() => ({
+  name: 'Switch',
+  icon: 'CheckboxIcon',
+  category: 'FORM',
+  structure: [
+    {
+      name: 'Checkbox',
+      options: [
+        {
```
flaviumariean, line 9: This is almost the same code as checkbox prefab excepting the "isSwitch" option. We can add this option to checkbox prefab and it should work fine without duplicating the code.

vdberganton:
Tested it and the styling options are not working because its lacking the correct MUI classname

src/prefabs/switch.js
vdberganton: From a coding perspective you're right. 

But with the no-coder in mind I think it is better to add multiple prefabs in the sidebar where they can choose their different "components" from.

flaviumariean (approved):

PR #788, opened 2021-01-13
Fix/autocomplete not querying more results

flaviumariean:


src/components/autoComplete.js
```diff
@@ -381,6 +401,7 @@
         freeSolo={freeSolo}
         autoSelect={freeSolo}
         options={results}
+        filterOptions={filterOptions}
```
chielwester, line 404: Are we now fetching the entire collection in multiple data api calls? That's not entirely what I expected here. If you remove the code on line 393: 

`
          if (!freeSolo) {
            return;
          }
`

The component will already behave as expected. Loading entire collections can also result in degraded performance when using it on large collections.

jeroenreijs (approved):

PR #792, opened 2021-01-14
Feat/update dialog interactions to new names and box

MarcelKorporaal:
Dialog Prefab had old DialogClose interaction in it which was renamed to 'Hide' so I updated it.
Also the box descendant inside the dialog prefab was an older version with less options than the current box so I added these options from the Box prefab


MarcelKorporaal:
Closed due incomplete update. There's another box in the prefab which needs to be updated

PR #799, opened 2021-01-18
chore: add remove function to box component

DylanCarver:
Improve function naming & add remove function to box component

### DIFF
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -31,7 +31,7 @@
     });
 
     B.defineFunction('removeCustomBackgroundImage', () => {
-      setInteractionBackground(false);
+      setInteractionBackground('');
     });
 
     const boxOptions = {
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -26,8 +26,12 @@
     const [opacity, setOpacity] = useState(opac);
     const [interactionBackground, setInteractionBackground] = useState('');
 
-    B.defineFunction('setBackgroundImage', url => {
-      setInteractionBackground(url);
+    B.defineFunction('setCustomBackgroundImage', url => {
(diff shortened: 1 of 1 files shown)


src/components/box.js
```diff
@@ -26,8 +26,12 @@
...
     const [interactionBackground, setInteractionBackground] = useState('');
 
-    B.defineFunction('setBackgroundImage', url => {
-      setInteractionBackground(url);
+    B.defineFunction('setCustomBackgroundImage', url => {
+      setInteractionBackground(`url("${url}")`);
+    });
+
+    B.defineFunction('removeCustomBackgroundImage', () => {
+      setInteractionBackground(false);
```
BNNBetty, line 34: Keep types consistent for this state value, maybe reset it to empt6y string since that is the default

src/components/box.js
```diff
@@ -63,9 +67,7 @@
         onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
         style={{
-          backgroundImage: interactionBackground
-            ? `url("${interactionBackground}")`
-            : 'auto',
+          backgroundImage: interactionBackground || hasBackgroundImage,
```
BNNBetty, line 70: Here the backgroundImage is either a string or a boolean which is also a bit inconsistent. maybe find a way to check the desired states but provide only one type of value to the prop itself to avoid polymorphism down the line

DylanCarver: I've refactored this a bit. Now when an interaction is used it uses the inline styling to apply it, otherwise, no inline styling is applied, and the optional background is handled by the `styles` section. This also cleans up the types.

BNNBetty (approved):

PR #808, opened 2021-01-19
feat: added link option to an image

Reinier1982:
The image of the media component has the option now to add a link to it

### DIFF
--- a/src/components/media.js
+++ b/src/components/media.js
@@ -106,12 +106,8 @@
       MediaComponent = () => (
         <Link
           href={href}
-          component={
-            linkType === 'internal' && hasInteralLink ? B.Link : undefined
-          }
-          endpoint={
-            linkType === 'internal' && hasInteralLink ? linkTo : undefined
-          }
+          component={hasInteralLink ? B.Link : undefined}
+          endpoint={hasInteralLink ? linkTo : undefined}
         >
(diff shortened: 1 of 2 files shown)


src/components/media.js
```diff
@@ -28,6 +34,15 @@
     const variable = imageSource && imageSource.findIndex(v => v.name) !== -1;
     const variableDev = env === 'dev' && (variable || !imgUrl);
 
+    const hasLink = linkTo && linkTo.id !== '';
```
jeroenreijs, line 37: I would write the following
```
    const hasLink = linkTo && linkTo.id !== '';
    const hasExternalLink = linkToExternal && linkToExternal.id !== '';
    const anyLink = hasLink || hasExternalLink;
```
as this:
```
    const hasInternalLink = linkTo && linkTo.id !== '';
    const hasExternalLink = linkToExternal && linkToExternal.id !== '';
    const hasLink = hasInternalLink || hasExternalLink;
```
Variables with names starting with `is` and `has` mean they contain a boolean value. When it starts with `any`, it gives another idea. Remember that the current `anyLink` is used in the other code, so that has to be renamed as well.

jeroenreijs (changes requested):
See my comment...

jeroenreijs (approved):

PR #817, opened 2021-01-20
Fix/clear file upload interaction

chielwester:


src/components/fileUpload.js
```diff
@@ -71,7 +71,7 @@
     };
 
     const clearFiles = e => {
-      e.preventDefault();
+      if (e && e.preventDefault()) e.preventDefault();
```
jeroenreijs, line 74: You probably just want to know if `e.preventDefault` is defined, not to test the result of the function.

chielwester: Fixed that!

jeroenreijs (approved):

PR #823, opened 2021-01-21
Feat/add onchange option to select

Reinier1982:
The default Change trigger doesn't work for select boxes. Added it's own one as discussed with Martijn Jansen.

### DIFF
--- a/src/components/media.js
+++ b/src/components/media.js
@@ -5,6 +5,7 @@
   orientation: 'VERTICAL',
   jsx: (() => {
     const { env, useText } = B;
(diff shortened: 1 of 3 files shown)


jeroenreijs (changes requested):
Please create 2 PR's for this obviously different things. That way we can handle them separately from eachother and they will both be present in the release notes.

jeroenreijs:
I notice this PR as well: https://github.com/bettyblocks/material-ui-component-set/pull/808 which contains one of these commits. So just create a new branch instead of using the same for adding 2 different changes please.

PR #824, opened 2021-01-21
fix: fix optiontype setting for list property in prefab

garethloot:
Screen Shot 2021-01-21 at 19 03 31

When list property was selected in beforeCreate wizard property type was set incorrectly in prefab


jeroenreijs:
Why don't you create a branch in the repo itself? Makes merging easier and you have access. Also, when the type of your commit is a fix, your message doesn't need to start with that. Now it says `fix: fix ....`.

PR #848, opened 2021-01-28
feat: drawer menu to toggle with buttons

Reinier1982:


Reinier1982:
The current drawer in the PB, has 3 elements which are defined in the prefab of the drawer, which makes it's hard to change it to only be the menu part of the drawer. This is why it's a separate component.

jeroenreijs (approved):

jeroenreijs:
https://bettyblocks.atlassian.net/browse/TA-2264

PR #852, opened 2021-01-29
fix: useFilter with _or fails when using search in data table

chielwester:
When you use a filter with multiple rows and OR, and also apply a search property in the data table the useFilter helper fails resulting in an error `Property in filter not found`. This change fixes that by first resolving both the filter and search property filter via useFilter and then combining them.

### DIFF
--- a/src/components/dataTable.js
+++ b/src/components/dataTable.js
@@ -147,12 +147,15 @@
         )
       : {};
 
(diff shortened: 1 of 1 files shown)


chielwester:
Not needed anymore, already fixed at platform level

PR #860, opened 2021-02-02
Feat/conditional table columns and triggers{3022}

garethloot:
Fix missing initial visibility toggle for data Columns


garethloot:
Doing it direct to betty repository

PR #885, opened 2021-02-08
Feat/drawer menu to toggle with buttons

Reinier1982:


jeroenreijs:
Why did you remove the media prefab and component? This was probably by accident?

jeroenreijs (changes requested):
See my comment about the removed media prefab and component. Please undo that.

PR #890, opened 2021-02-09
Feat/box backgroundattachment

robinvanderburght:
With this option you have the ability to define if you want to have a static background image or not. In this you you can prevent the background image from 'jumping' when pages have different heights or tabs within the page have different heights.


jeroenreijs:
We will pick this up with this ticket: https://bettyblocks.atlassian.net/browse/TA-2283

PR #903, opened 2021-02-15
Fix(ART-2986): Resolve ModelSelector implementation to only receive the UUID

roywillemse:
Resolve ModelSelector component so the modelObject is no longer needed and the model is fetched in the prefab itself.

### DIFF
--- a/src/prefabs/dataContainer.js
+++ b/src/prefabs/dataContainer.js
@@ -22,8 +22,9 @@
   }) => {
     const [anotherPageState, setAnotherPageState] = React.useState({
       modelId: '',
-      model: null,
     });
+
+    const [modelId, setModelId] = React.useState(null);
 
     const [loggedInUserState, setLoggedInUserState] = React.useState({
       authenticationProfile: null,
@@ -31,7 +32,6 @@
 
     const [thisPageState, setThisPageState] = React.useState({
       modelId: null,
-      model: null,
       component: null,
     });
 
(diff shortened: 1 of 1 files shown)


src/prefabs/dataContainer.js
```diff
@@ -25,6 +25,8 @@
       model: null,
     });
 
+    const [fetchModelId, setFetchModelId] = React.useState(null);
```
stefan-betty, line 28: I think we should rename it to `fetchedModelId`. The name `fetchModelId` give me the indication that it still needs to be fetched, which is not true.

roywillemse: At the moment of setting the `fetchModelId` it does still need to be fetched right? I am using this state variable so we can always use this in the ModelQuery

roywillemse: `fetchModelData` is maby a better name? Because we are using the UUID in this state to fetch the model's data. Agreed?

stefan-betty: I've we still have to fetch it, I think `fetchModelId` will be fine!

src/prefabs/dataContainer.js
```diff
@@ -51,10 +53,22 @@
...
       if (data) {
-        setThisPageState(prevState => ({
-          ...prevState,
-          model: data.model,
-        }));
+        switch (buttonGroupValue) {
+          case 'anotherPage':
+            setAnotherPageState(prevState => ({
+              ...prevState,
+              model: data.model,
```
emielwit, line 60: Do you still need to store the model in `state`? `data` is updating itself when the`model_id` changed

src/prefabs/dataContainer.js
emielwit, line 28: I prefer to name this `[modelId, setModelId]`

PR #904, opened 2021-02-15
Added redirect option to button on Action completed

robinvanderburght:
This adds a redirect option on Action completed to the button component.

Reason is that sometimes you just want to update a status (or something like that) and afterwards redirect to another page, without the need to pass all kinds of input variables like in af from component. I've got this feedback the last 2 months multiple times from self building customers.

### DIFF
--- a/src/components/button.js
+++ b/src/components/button.js
@@ -23,6 +23,7 @@
       actionId,
       buttonText,
       actionModels,
+      redirect,
     } = options;
     const {
       env,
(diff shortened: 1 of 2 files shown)


jeroenreijs:
We will pick this up with ticket https://bettyblocks.atlassian.net/browse/TA-2303

BarryBlocks:
@robinvanderburght This can be accomplished via the redirect interaction in combination with the 'onActionSucces' trigger. Therefore, I'll close this ticket.

PR #924, opened 2021-02-16
add interactions and triggers {TECHSUP-3287}

garethloot:
- Add isTrue, isFalse and onChange triggers to switch and checkbox
- Add set visibility interaction to conditional
- Add onChange trigger to radio group


garethloot:
For demo look at https://form-demo.betty.app/waterfall

flaviumariean (approved):

PR #968, opened 2021-03-01
Testing data tags

garethloot:
Add an advanced toggle option that shows een text field where you can type a value that is put in a data tag. This can then be used as target in testing software to select the component. This option has been added to content components and input for forms.

techsub-3375

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,17449 @@
+{
+  "name": "component-set",
+  "version": "1.107.1",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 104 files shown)


BarryBlocks:
There's a techsup ticket created for this (https://bettyblocks.atlassian.net/browse/TECHSUP-3385). Therefore, any further actions will follow once this techsup is linked to a team, so they will look over the proposed changes.

vdberganton:
@garethloot  closing this because it was fixed and merged to production via another branch

PR #1007, opened 2021-03-18
feat: grid column creator

Reinier1982:


src/prefabs/columnLayout.js
```diff
@@ -0,0 +1,469 @@
...
+                }
+              >
+                <Button
+                  label="+ Add Row"
+                  disabled={!maxRows}
+                  onClick={() => {
+                    if (maxRows) {
+                      setRows([
+                        ...rows,
+                        { row: rows.length + 1, columns: '1' },
```
jeroenreijs, line 70: This key `row` should be `index`.

jeroenreijs (changes requested):

jeroenreijs (approved):

PR #1027, opened 2021-04-01
fix: only render children when datacontainer has data

chielwester:
Prevents the children of a datacontainer to be rendered when the data container itself is still loading its data from an authentication profile. This caused input components that use a state for the default value (like the auto complete and hidden input) to incorrectly initialize the state with an empty value.

### DIFF
--- a/src/components/dataContainer.js
+++ b/src/components/dataContainer.js
@@ -153,9 +153,12 @@
             B.triggerEvent('onNoUserResults');
           }
 
+          if (!data) {
+            return <div />;
+          }
           return (
             <MeProvider value={data} id={model}>
-              <CanvasLayout />
+              {children}
             </MeProvider>
           );
         };


BarryBlocks:
I've added a ticket to our backlog (https://bettyblocks.atlassian.net/browse/TA-2467). Further progress can be followed there.

src/components/dataContainer.js
```diff
@@ -153,9 +153,12 @@
             B.triggerEvent('onNoUserResults');
           }
 
+          if (!data) {
+            return <div />;
+          }
           return (
             <MeProvider value={data} id={model}>
-              <CanvasLayout />
+              {children}
```
jeroenreijs, line 161: What do we need this change? `CanvasLayout` is either `BuilderLayout` or `One`, which both render children.

BarryBlocks:
@chielwester Could you elaborate on Jeroen's question in the comments?

vdberganton:
Close due to inactivity

PR #1039, opened 2021-04-08
feat: page template with login and register form TMPLT-368

RobD-tech:
Added pageWithLoginAndRegisterForm.js


jeroenreijs (changes requested):
`Type: 'page'` is commented, so this won't appear in the page templates overview.

Line 5854 and 7948, could't the iconConfiguration be used here? Saves a lot of lines once more.

This doesn't look OK:
image

Also I noticed the success message when using the register form but didn't fill in any field, is that OK? 
image

jeroenreijs:
The strange looking properties table is because of missing translations, that is not something that you can fix. So please ignore the comment `This doesn't look OK:`

jeroenreijs (approved):

PR #1045, opened 2021-04-09
feat/page-with-create-datatable-tmplt-421

sventruschel:
Added datatable with create TMPLT-421

### DIFF
--- a/src/prefabs/pageWithCreateDataTable.js
+++ b/src/prefabs/pageWithCreateDataTable.js
@@ -41,7 +41,7 @@
       skip: !modelId,
     });
 
-    const reduceStructure = (refValue, structure) =>
+    const getDescendantByRef = (refValue, structure) =>
       structure.reduce((acc, component) => {
         if (acc) return acc;
         if (
@@ -51,7 +51,7 @@
         ) {
           return component;
         }
-        return reduceStructure(refValue, component.descendants);
+        return getDescendantByRef(refValue, component.descendants);
       }, null);
 
     const iconConfiguration = {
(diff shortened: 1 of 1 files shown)


src/prefabs/pageWithCreateDataTable.js
```diff
@@ -0,0 +1,12336 @@
...
+                'SIGNED_PDF',
+                'SUM',
+                'BOOLEAN_EXPRESSION',
+                'DATE_EXPRESSION',
+                'DATE_TIME_EXPRESSION',
+                'DECIMAL_EXPRESSION',
+                'INTEGER_EXPRESSION',
+                'MINUTES_EXPRESSION',
+                'PRICE_EXPRESSION',
+
```
jeroenreijs, line 6556: It's better to define this list just once and use it at both PropertiesSelectors.

jeroenreijs, line 9324: Your new `autoOK` option needs to be added here as well

jeroenreijs, line 9652: Your new `autoOK` option needs to be added here as well

jeroenreijs, line 10016: Your new `autoOK` option needs to be added here as well

jeroenreijs (changes requested):

src/prefabs/pageWithCreateDataTable.js
```diff
@@ -0,0 +1,12336 @@
...
+      createFormUseDataProperties,
+      setCreateFormUseDataProperties,
+    ] = React.useState(true);
+
+    const { data } = useModelQuery({
+      variables: { id: modelId },
+      skip: !modelId,
+    });
+
+    const reduceStructure = (refValue, structure) =>
```
jeroenreijs, line 44: This is just a suggestion. it's not wrong like it is right now, see the other PR https://github.com/bettyblocks/material-ui-component-set/pull/1048. But it looks like a lookup for a certain prefab in the structure, though the function name is rather global. Maybe you can think of something more descriptive.

src/prefabs/pageWithCreateDataTable.js
sventruschel: Defined it once and used it in both properties selectors

sventruschel: Added autoOK option

sventruschel: Added autoOK option

src/prefabs/pageWithCreateDataTable.js
```diff
@@ -0,0 +1,12156 @@
...
+                />
+              </Field>
+            )}
+          </Field>
+        </Content>
+        <Footer
+          onSave={() => {
+            if (
+              !modelId ||
+              (createFormProperties.length < 1 &&
```
nulian, line 6405: Why is length < 1 being done here. Feels like a weird way to check if array is empty.

The above `!createFormProperties.length` seems like a better way to check it.

Same for the cases below this.

nulian (changes requested):

src/prefabs/pageWithCreateDataTable.js
prorobbert: That's a good feedback point. I'll try to keep using the shorter method from now on.

PR #1046, opened 2021-04-09
feat: enable auto close on select - TMPLT-168

RobD-tech:
Enable the option to auto-close the picker modal after a value is selected. Implemented for the datePicker, dateTimePicker and timePicker. Default value is true.

### DIFF
--- a/src/components/dateTimePicker.js
+++ b/src/components/dateTimePicker.js
@@ -19,6 +19,7 @@
       margin,
       helperText,
       disableToolbar,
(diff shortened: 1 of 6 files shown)


jeroenreijs (changes requested):
DateTimePicker is also used in `CreateForm` and `UpdateForm`. Please add the new option in those prefabs as well.

vdberganton:
Closing due to inactivity

PR #1048, opened 2021-04-12
feat: page template with edit data table - TMPLT-383

RobD-tech:
Page template containing a data table and an edit form within a dialog.


src/prefabs/pageWithEditDataTable.js
```diff
@@ -0,0 +1,11725 @@
...
+                          },
+                        },
+                        {
+                          value: false,
+                          label: 'Hide label',
+                          key: 'hideLabel',
+                          type: 'TOGGLE',
+                          configuration: {
+                            condition: {
+
```
jeroenreijs, line 8668: Your new `autoOK` option needs to be added here as well

jeroenreijs, line 8991: Your new `autoOK` option needs to be added here as well

jeroenreijs, line 9353: Your new `autoOK` option needs to be added here as well

src/prefabs/pageWithEditDataTable.js
```diff
@@ -0,0 +1,11725 @@
...
+    ] = React.useState(false);
+    const [editProperties, setEditProperties] = React.useState([]);
+    const [inheritProperties, setInheritProperties] = React.useState(true);
+
+    const { data } = useModelQuery({
+      variables: { id: modelId },
+      skip: !modelId,
+    });
+
+    const reduceStructure = (refValue, structure) =>
```
jeroenreijs, line 42: This is just a suggestion, but since this is used to find a certain prefab in the structure, maybe it's better to call it something else than `reduceStructure`. I noticed it in the other PR as well. https://github.com/bettyblocks/material-ui-component-set/pull/1045. It's not wrong, but doesn't give any info either.

jeroenreijs (changes requested):

src/prefabs/pageWithEditDataTable.js
RobD-tech: Added `autoOk` option

RobD-tech: Added `autoOk` option

RobD-tech: Added `autoOk` option

src/prefabs/pageWithEditDataTable.js
RobD-tech: Renamed `reduceStructure` to `getDescendantByRef`

jeroenreijs (approved):

jeroenreijs:
I've approved this, the appbar still has 60px height, I'll leave that up to you. Because there are no buttons in there, it doesn't cause an extra scrollbar for this situation.

PR #1052, opened 2021-04-14
feat: page template with master detail view vertically oriented - TMPLT-404

RobD-tech:
Page template containing a data table and detail view, vertically oriented.

### DIFF
--- a/src/prefabs/pageWithMasterDetailVertical.js
+++ b/src/prefabs/pageWithMasterDetailVertical.js
@@ -1725,7 +1725,7 @@
                     type: 'SIZE',
                     label: 'Height',
                     key: 'height',
-                    value: '60px',
+                    value: '',
                     configuration: {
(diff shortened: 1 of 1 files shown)


jeroenreijs (approved):

src/prefabs/pageWithMasterDetailVertical.js
```diff
@@ -0,0 +1,4035 @@
...
+              onChange={value => {
+                setDetailsValidation(!value.length);
+                setDetailProperties(value);
+              }}
+            />
+          </Field>
+        );
+      },
+      onSave: () => {
+        if (!modelId || properties.length < 1 || detailProperties.length < 1) {
```
nulian, line 1696: To me it seems better to check with like !properties.length and probably assign it before the if.
const hasProperties
const hasDetailProperties

Because it is also used multiple times.

PR #1059, opened 2021-04-19
added striped option

robinvanderburght:
A feature request from multiple self-building (consultancy) customers was the option to have striped DataTables. An option was added to make the table striped and also an option to select the color.

### DIFF
--- a/src/components/dataTable.js
+++ b/src/components/dataTable.js
@@ -44,6 +44,7 @@
       labelNumberOfPages,
       labelSearchOn,
       square,
+      striped,
       elevation,
       variant,
       stickyHeader,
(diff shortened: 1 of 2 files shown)


src/components/dataTable.js
```diff
@@ -251,6 +252,22 @@
           repeaterRef.current.innerHTML +=
             repeaterRef.current.previousElementSibling.children[0].outerHTML;
         }
+        if (striped) {
```
jeroenreijs, line 255: When you make sure the CSS selector is selecting the right element in the autorepeat, you don't need this logic.
The selector `nth-child` doesn't look to the type of elements.
In the autorepeat we see children like this: ```<tr>...</tr><div>...</div><tr>...</tr><div>...</div><tr>...</tr><div>...</div>```
So what you need for this is probably a selector like this: ```      autoRepeat: {
        opacity: 0.5,
        '& > :nth-child(4n)': {
          backgroundColor: ({ options: { striped, stripeColor } }) => [
            striped ? style.getColor(stripeColor) : 'transparent',
          ],
        },
      },
```
And you might need `!important`. Not sure what the default is for Material UI.

jeroenreijs: Ok, this is impossible with the current setup, so the javascript solution is the only way.

jeroenreijs (approved):

PR #1071, opened 2021-04-21
Feat: page with grid view - TMPLT-442

RobD-tech:
Page template with grid view. Currently on edge, and tested on both edge and acceptance.


jeroenreijs (changes requested):
Remove the height of 60px please, it will cause an extra scrollbar.

jeroenreijs (approved):

PR #1089, opened 2021-04-25
Added option for numbered list

Anita-Heiblom:


src/components/dataList.js
```diff
@@ -48,6 +50,17 @@
         const isInline = type === 'inline';
         const isGrid = type === 'grid';
 
+        const NumberTag = {
```
jeroenreijs, line 53: Changing this option doesn't change anything in the output.

jeroenreijs (changes requested):
We also have a listWithData, which uses the datalist and some page templates (that might not be on acceptance yet). They should also have these extra options in order not to break.
But did you take a look at the Material UI docs about the List component?
You can use the `component` property to specify the HTML element you want it to render. And you can do that for ListItem as well. So maybe you can just use that propery, and change some CSS, to have a numbered list (you'll need to change `list-style` and `padding` at least). It's worth it looking into that.

vdberganton:
Closing due to inactivity

PR #1116, opened 2021-05-03
Feat/add style editor to button work 332

dylanBetty:


NiekBetty (approved):
Just as double check: I noticed in previous PR's that you changed the index when selecting options in the `beforeCreate` config of components. By changing the position of the icon option do these indexes need change too?

dylanBetty:
> Just as double check: I noticed in previous PR's that you changed the index when selecting options in the `beforeCreate` config of components. By changing the position of the icon option do these indexes need change too?

@NiekBetty This shouldn't affect the index as it's only switching around the order of the size and icon options.

PR #1142, opened 2021-05-11
Feat/add link option to image

prorobbert:


prorobbert:
The code has been merged. Github unfortunately doesn't see this

PR #1235, opened 2021-06-18
Added interactions to set lhs and rhs

Anita-Heiblom:


src/components/conditional.js
```diff
@@ -70,6 +74,24 @@
...
+        });
+        B.defineFunction('Set Right Value', value => {
+          if (typeof value === 'object') {
+            value.length == undefined
+              ? setRightValue(value.target.value)
+              : setRightValue(value.map(String));
+          } else {
+            setRightValue(value.toString());
+          }
+        });
```
vdberganton, line 94: You're doing exactly the same twice, try to prevent duplicate code by extracting the same logic in a separate function: 
```
const getValue = evt => {
          if (typeof evt !== 'object') {
            return evt.toString();
          }
          if (evt.length !== undefined) {
            return evt.map(String);
          }
          if (evt.target && !evt.target.checked) {
            return [];
          }
          return evt.target && evt.target.value;
        };

        B.defineFunction('Set Left Value', event =>
          setLeftValue(getValue(event)),
        );
        B.defineFunction('Set Right Value', event =>
          setRightValue(getValue(event)),
        );
```

Also you can use `yarn lint` after making your changes to check if you have any errors before you open a PR and do this via the pipeline.

vdberganton, line 94: Always try to prevent duplicate logic by extracting that and putting it in a separate function.

```
const getValue = evt => {
  if (typeof evt !== 'object') {
    return evt.toString();
  }
  if (evt.length !== undefined) {
    return evt.map(String);
  }
  if (evt.target && !evt.target.checked) {
    return [];
  }
  return evt.target && evt.target.value;
};

B.defineFunction('Set Left Value', event =>
  setLeftValue(getValue(event)),
);
B.defineFunction('Set Right Value', event =>
  setRightValue(getValue(event)),
);
```

Also, you have linter errors, you can check these before you open a pull request by putting `yarn lint` in your terminal. 

I can't merge this in its current state and need to investigate why some form inputs only send values on changes.

PR #1251, opened 2021-06-25
fix form submit interaction for Safari

peterbeers:


jeroenreijs (changes requested):
This would mean that when a browser supports both, the form is submitted twice. Also, we'd like you to use braces when a condition is over more than one line.

jeroenreijs (approved):

PR #1261, opened 2021-06-30
Feat/add on change filter support for datetime page 322

dehlak:


src/components/dateTimePicker.js
```diff
@@ -67,6 +67,7 @@
     const isValidDate = date => date instanceof Date && !isNaN(date);
 
     const changeHandler = date => {
+      B.triggerEvent('OnChange', date);
```
benjamin-rocks, line 70: lower case 'o'

tee hee

benjamin-rocks (approved):

PR #1323, opened 2021-07-30
Added min/max date

Anita-Heiblom:
Added option an interaction to datepicker to set min and max date or min days before/max days ahead

### DIFF
--- a/src/components/dateTimePicker.js
+++ b/src/components/dateTimePicker.js
@@ -14,6 +14,9 @@
       dateFormat,
       timeFormat,
       dateTimeFormat,
+      useMinMaxDate,
+      minDaysBefore,
+      maxDaysAhead,
       size,
       fullWidth,
       margin,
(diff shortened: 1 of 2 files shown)


vdberganton:
Created: https://bettyblocks.atlassian.net/browse/PAGE-678

src/components/dateTimePicker.js
```diff
@@ -41,6 +44,52 @@
     const [selectedDate, setSelectedDate] = useState(null);
     const helper = useText(helperText);
     const placeholderText = useText(placeholder);
+    const parsedMinDate =
```
jeroenreijs, line 47: Using a `let` instead of a `const` might be better here. This way you can avoid disabling the linter. Something like this would do:

```
let parsedMinDate;
let parsedMaxDate;
if (useMinMaxDate === 'date') {
  parsedMinDate = new Date(useText(options.minDate));
  parsedMaxDate = new Date(useText(options.maxDate));
} else if (useMinMaxDate === 'days') {
  parsedMinDate = useText(minDaysBefore);
  parsedMaxDate = useText(maxDaysAhead);
}
```
There is nothing wrong with using a `let`. And it will make this code easier and saves some lines.

src/prefabs/datePicker.js
```diff
@@ -56,6 +56,77 @@
           key: 'dateFormat',
           type: 'TEXT',
         },
+        {
+          label: 'Use min/max date',
```
jeroenreijs, line 60: When the datePicker is used in other prefabs, these options might need to be added in those prefabs as well.

src/components/dateTimePicker.js
```diff
@@ -41,6 +44,52 @@
...
+    const parsedMaxDate =
+      // eslint-disable-next-line no-nested-ternary
+      useMinMaxDate === 'date'
+        ? new Date(useText(options.maxDate))
+        : useMinMaxDate === 'days'
+        ? useText(maxDaysAhead)
+        : undefined;
+
+    const getMinDate = value => {
+      if (!value) return undefined;
```
jeroenreijs, line 63: When you extend this condition with the fact that `useMinMaxDate` might not be defined or has `none` as value, you can have even simpler conditions.

Something like this: 
```
const hasDateMinMax = useMinMaxDate && useMinMaxDate !== 'none';

const getMinDate = value => {
  if (!value || !hasDateMinMax) return undefined;
  if (useMinMaxDate === 'date') return value;
  const newMinDate = new Date();
  return newMinDate.setDate(newMinDate.getDate() - parseInt(value, 10));
};
```
And of course for the `getMaxDate` as well.

jeroenreijs (changes requested):

vdberganton:
Closing due to inactivity

PR #1330, opened 2021-08-05
Feat/autocomplete option

ihadjo:
Add autocomplete option to:
DatePicker, DateTime, emailInput, decimalInput, numberInput, priceInput, textField and forms


jeroenreijs:
Are you sure about this? This looks like edge has been merged to this branch somewhere along the way. Maybe to solve conflicts?

vdberganton:
I think you're right Jeroen, I will see the actual changes that were done and create a new PR with only those changes.

PR #1347, opened 2021-08-09
fix: remove condition query hooks

benjamin-rocks:


src/components/dataTable.js
```diff
@@ -224,27 +224,25 @@
...
-        onError(err) {
-          if (!displayError) {
-            B.triggerEvent('onError', err);
-          }
-        },
-      });
+    const { loading, error, data, refetch } = useAllQuery(model, {
+      rawFilter: where,
+      variables,
+      skip: !model || loadOnScroll ? skip : page * rowsPerPage,
```
vdberganton, line 230: skip: loadOnScroll ? skip : (page ? (page - 1) * rowsPerPage : 0),

Else the filter interaction does not show the record when on page 1

benjamin-rocks: Hey @vdberganton thanks, I'll make that change :)

vdberganton (approved):

PR #1361, opened 2021-08-17
fix: prevent onChange of textfield to be triggered on load

chielwester:
The Textfield component triggers an onSave automatically on load with a default value which results in unwanted behaviour. For example if you have auto-submit based on changes of the textfield the save is already triggered as soon as the text field loads. This PR prevents that, functionality is based on the same implementation that is already used in the select component

### DIFF
--- a/src/components/textField.js
+++ b/src/components/textField.js
@@ -72,6 +72,8 @@
       customModelAttributeId,
     );
 
+    const mounted = useRef(false);
+
     const {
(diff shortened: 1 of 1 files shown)


jeroenreijs:
I'm closing this PR because we had a duplicate https://github.com/bettyblocks/material-ui-component-set/pull/1360. In this one (1360) I liked the chosen solution better.

PR #1374, opened 2021-08-19
fix: login interaction uses react router

garethloot:
Make sure the login interaction uses react router and not window location href. This prevents a complete reload of the application.


src/interactions/login.ts
```diff
@@ -22,6 +26,6 @@ function login({
   if (isValid) {
     localStorage.setItem('TOKEN', jwtToken);
     localStorage.setItem('REFRESH_TOKEN', refreshToken);
-    window.location.href = url;
+    history.push(url);
```
benjamin-rocks, line 29: One suggestion could you change history to `window.history` and remove the interface?

benjamin-rocks: never mind. I see this is a global instance of react router

benjamin-rocks (approved):

PR #1382, opened 2021-08-23
feat: filter value send in string format of slider

Angelodaniel:


SanderVanRiessen (approved):

jeroenreijs (changes requested):
There is a prop `onChange` on the slider. Do you think it's possible to use that instead of an extra useEffect?

jeroenreijs:
Change no longer needed

PR #1427, opened 2021-09-09
Fix/use local datetime in datepicker page 830

jeroenreijs:


jeroenreijs:
We won't fix this in the component. This is not sufficient and should be handled in the Data API

PR #1520, opened 2021-10-28
feat/tabs-load-on-show/pre-load-tabs

ihadjo:
Fix to reduce amount of call to data-api.


robbertruiter:
Why has this pull request been approved? The changes have not been implemented in the page template nor did we get a heads up about this. Currently our page templates using the tabs are broken

MenouerBetty:
Just talked to @ihadjo it appears he forgot to add the option into the page prefabs as well

PR #1551, opened 2021-11-04
fix: text field doesn’t throw hook errors on validation errors

chielwester:
TextField component threw errors when using dynamic values in combination with validations because the useText helper was used in a function. This resulted in invalid hook errors causing the component to block from further input. This change fixes it by using the useText hook in the proper way


jeroenreijs:
We handled all form field components ourselves today. Thanks for the input. But this PR is no longer needed and will be closed.

PR #1731, opened 2022-02-17
Feat/set rows per page through button pfr460

peterbeers:


src/components/button.js
```diff
@@ -222,6 +222,10 @@
 
     const handleClick = (e) => {
       e.stopPropagation();
+
+      if (!isNaN(buttonText)) {
+        B.triggerEvent('sendRowsPerPage', buttonText);
```
M-Ryan92: typo? should be `setRowsPerPage` right?

src/prefabs/button.js
```diff
@@ -22,6 +22,12 @@
           key: 'buttonText',
           value: ['Button'],
         },
+        {
+          type: 'VARIABLE',
+          label: 'Button value',
+          key: 'buttonValue',
+          value: [''],
+        },
```
M-Ryan92, line 30: please include this option to all the prefabs where an button is used. :pray:

M-Ryan92 (approved):

PR #1735, opened 2022-02-18
Feat/hide rows per page pfr464

peterbeers:


src/prefabs/dataTable.js
```diff
@@ -343,6 +343,12 @@
             },
           },
         },
+        {
```
dylanBetty, line 346: @peterbeers This prefab change should be applied to all prefabs that contain the dataTable component, so that the component stays consistent over all prefabs and page templates, unfortunately this isn't inherited at the moment.

The dataTable component is listed in the following prefabs:
- [x] Datatable
- [ ] BackOffice
- [ ] pageWithCreateDatatable
- [ ] pageWithEditDatatable
- [ ] pageWithCrudSlideOutPanel (x3)
- [ ] pageWithOverviewAndDetails
- [ ] pageWithSimpleDashboard

dylanBetty (approved):

PR #1737, opened 2022-02-21
Fix/form callback not triggered edge

roywillemse:
Because the defineFunction and the triggerEvent were both in a separate useEffect, they would get initialised at the same time which caused the callback to not be executed sometimes.

@jeroenreijs Hopefully you want to review this for me :)
Only change inside: `src/components/form.js`
I have moved the onComponentRendered from the separate useEffect into the same useEffect as the functions are defined, because in separate useEffects the trigger would happen before the defineFunction was ready.


roywillemse:
Closing because of changelog changes in the PR that shouldn't be there

PR #1757, opened 2022-03-08
new setValue interaction

robinvanderburght:
In some specific cases you want to set the value from another textfield based on your input. This simple interaction covers those use cases.


benjamin-rocks (approved):

benjamin-rocks (changes requested):
You may want to do some validation

benjamin-rocks (approved):

PR #1759, opened 2022-03-09
feat: add show development step option

gijsvanwichen:
Example: 
Your steps component has 5 steps. Step 1 opens on page load, but you got feedback on step 3.

Current situation: 
1. Navigate to steps component en set the 'Show step' on step 3.
2. Make changes on step 3
3. Navigate to steps component en set the 'Show step' on step 1. (This step is comon to forget, so after you merge it to the next environment you'll receive feedback that the steps component opens on step 3 instead on step 1)
4. Compile page
5. Click on step 3 en test your changes

New situation:
1. Navigate to steps component en set the 'Show development step' on step 3.
2. Make changes on step 3
3. Compile page
4. Click on step 3 en test your changes

The new situation improves the development speed, the developer happiness and les unnecessary feedback.

### DIFF
--- a/src/prefabs/stepper.js
+++ b/src/prefabs/stepper.js
@@ -15,8 +15,8 @@
         },
         {
           type: 'NUMBER',
-          label: 'Show development step',
-          key: 'activeDevelopmentStep',
+          label: 'Selected design step index',
+          key: 'selectedDesignStepIndex',
           value: '1',
         },
         {
--- a/src/components/stepper.js
+++ b/src/components/stepper.js
@@ -16,7 +16,7 @@
     const { env, useText, Children, Icon } = B;
     const {
       activeStep: stepIndex,
-      activeDevelopmentStep,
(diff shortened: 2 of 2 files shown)


src/prefabs/stepper.js
```diff
@@ -13,6 +13,12 @@
           key: 'activeStep',
           value: '1',
         },
+        {
```
dylanBetty, line 16: Maybe this could go under the advanced settings options, what do you think?

dylanBetty: Also I would align the label and key to be the same for both the stepper and tabs component to keep consistency.

vdberganton: Don't think it should be under advanced as its probably one of the most used options.

But please rename it to "Selected design tab index" as we communicate working in the Page Builder officially as "Design mode"

dylanBetty (approved):

PR #1783, opened 2022-03-24
feat: add enable and disable interaction on the checkboxgroup component

SanderVanRiessen:


BettyBrian (approved):
I've tested the checkbox group component and it works as intended. 
I've also pulled it through a 'check' and nothing seemed to be breaking. it passed everything

PR #1792, opened 2022-03-31
Feat/expansion panel prevent render childs

ihadjo:
https://bettyblocks.atlassian.net/browse/PFR-487

### DIFF
--- a/src/components/expansionPanel.js
+++ b/src/components/expansionPanel.js
@@ -84,7 +84,7 @@
       expandIcon: <Icon name="ExpandMore" />,
     };
 
-    const ShowChilderen = <div>{dataShown || isDev ? children : ''}</div>;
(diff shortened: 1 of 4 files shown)


src/components/expansionPanel.js
```diff
@@ -82,6 +84,8 @@
       expandIcon: <Icon name="ExpandMore" />,
     };
 
+    const ShowChilderen = <>{dataShown || isDev ? children : ''}</>;
```
dylanBetty, line 87: Are the fragments necessary here?

vdberganton (approved):

src/components/expansionPanel.js
ihadjo: I think so, otherwise I had a validation error

PR #1810, opened 2022-04-11
Feat/disable head stepper

robinvanderburght:
The stepper component currently misses the option to disable the navigation. In this way you can use it in a more flexible way. We see that a lot of self builders want to use the 'steps' on different pages (which is also better for the optimisation of the page in the IDE).


sventruschel:
If this is still something you want implemented, the prefab have been converted to tsx. So the option should be added there

PR #1838, opened 2022-04-25
feat: add datetime picker beta

benjamin-rocks:


src/components/actionJSForm.js
```diff
@@ -16,26 +16,12 @@
...
-          if (!error || !error.messages || !error.messages.errors) return [];
-          return error.message.errors.map((inner) => inner.message);
-        });
-
-        B.triggerEvent('onActionError', new Error(messages.join(', ')));
-
-        console.log('Errors: ', messages.join('\n   '));
-        return;
-      }
-
```
dylanBetty, line 31: Was the removal of the error handling removed deliberately? Is it correct that this is handled via the alert component now? As I see there have been some additions regarding error messages there.

benjamin-rocks: Yes, the error handling has been moved there. 

Ideally this should be done inside a helper

dylanBetty (approved):

src/components/actionJSForm.js
benjamin-rocks: Created issue PAGE-1845 for the helper

dylanBetty: :ok_hand:

PR #1842, opened 2022-04-26
Feat/date picker page 1557

benjamin-rocks:


src/prefabs/structures/DateTimePicker/options.ts
```diff
@@ -0,0 +1,178 @@
...
+    'Margin',
+    [
+      ['None', 'none'],
+      ['Dense', 'dense'],
+      ['Normal', 'normal'],
+    ],
+    { value: 'normal' },
+  ),
+
+  styles: toggle('Styles', { value: false }),
```
dylanBetty, line 110: In other components we split up the advanced / styles / validation files separately to reduce the size of this file. Might want to do the same here.

dylanBetty (approved):

PR #1849, opened 2022-04-26
fix: linter warnings

NiekBetty:
This PR:

- Enforces 0 linter warnings in the pipelines
- Adds a linter config for TypeScript files


src/interactions/logout.ts
```diff
@@ -1,23 +1,16 @@
+/* eslint @typescript-eslint/no-unused-vars: ["error", { "varsIgnorePattern": "logout" }] */
```
NiekBetty, line 1: What do you people think about this solution? I didn't want to completely disable the rule, but made an exception that needs to be manually added for every interaction.

benjamin-rocks: https://stackoverflow.com/questions/41316551/eslint-ignore-specific-rule-for-a-specific-directory

This might be easier in the long run, since we'd only do it once.

dylanBetty (approved):

src/interactions/logout.ts
NiekBetty: The thing is that we don't want to disable the entire rule. In some of the files there were true unused variables for which eslint correctly showed an error.

benjamin-rocks: understood

PR #1870, opened 2022-05-03
Rename autocomplete & remove validation related options

dylanBetty:


src/prefabs/structures/AutocompleteInput/options/validation.ts
```diff
@@ -0,0 +1,19 @@
...
+import { showOn } from '../../../../utils';
+
+export const validation = {
+  required: toggle('Required'),
+  validationValueMissing: variable('Value required message', {
+    value: ['This field is required'],
+    ...showOn('required'),
+  }),
+  closeOnSelect: toggle('Close dropdown after select', { value: true }),
+  disabled: toggle('Disabled'),
```
benjamin-rocks, line 12: toggles should have a valid value, does this show up in the lint?

dylanBetty: I've added an explicit value, it looks like the `attrs` object on the toggle function is optional 
```typescript
export declare const toggle: (label: string, attrs?: Attributes) => OptionProducer;
```

dylanBetty: Ah one moment, I see disabled has been put down twice, my linter didn't catch that for some reason.

benjamin-rocks: > ```export declare const toggle: (label: string, attrs?: Attributes) => OptionProducer;```

I thought i'd made it explicit, my mistake.

benjamin-rocks (approved):

PR #1977, opened 2022-06-02
feat(OL-915): added new property for templates where page type needs …

allamaley:
We, outlaws, have created a page type selector in NewPage create modal. We want to hide it for templates where authentication for the page does not make any sense.


jeroenreijs (changes requested):
You should also push the yarn.lock

jeroenreijs (approved):

PR #1987, opened 2022-06-07
chore: fixed the options in column prefabs

sventruschel:


BettyBrian (commented):
This PR will be overridden by a larger PR that will go to edge. That PR will include this work, therefore rendering this request unnecessary.

PR #1993, opened 2022-06-08
Feat/adding setvalue interaction inputs

Angelodaniel:
"I have a field in a form that would be used for a calculation.
This calculation must be directly executed en the result must be shown somewhere else.
It is not possible to use a form inside of a form, so I want to be able to sent my filled in value to the other input within the other form so I can submit that form and doe something with that calculation"

Example: 
In a webshop I can use the amount of products, I want to show the total price after the amount is known.


src/components/hiddenInput.js
```diff
@@ -38,6 +38,10 @@
       }
     }, [isDev, defaultValue]);
 
+    useEffect(() => {
+      B.defineFunction('setValue', (data) => setCurrentValue(data));
+    }, [currentValue]);
```
benjamin-rocks, line 43: we don't need `[currentValue]` as a dependency here.

src/components/textinput.js
```diff
@@ -183,6 +183,7 @@
     };
 
     B.defineFunction('Clear', () => setCurrentValue(''));
+    B.defineFunction('setInputValue', (data) => setCurrentValue(data));
```
benjamin-rocks, line 186: is there a reason why one is `setValue` and the other `setInputValue`.

benjamin-rocks (dismissed):

benjamin-rocks (approved):

PR #1998, opened 2022-06-09
Feat/adding setvalue interaction inputs

Angelodaniel:
Tested on Edge


sventruschel:
Looks like this branch is based on edge or really out of date. So if these changes are still necessary feel free to make a new PR

PR #2113, opened 2022-07-18
Z-index CSS fixed for sticky position option

robinvanderburght:
Also added the z-index option when the position is set to sticky. This gave the same bug as for the fixed position option


BettyBrian:
We're closing this request because this needs to be solved in the IDE

PR #2120, opened 2022-07-20
chore: allign naming with beforeCreate

jwielinga:


BettyBrian (changes requested):
When I wanna use the detail view prefab and use an image, it currently breaks the media component.
in the beforeCreate on Line 493 it checks if there is an image selected. If it is, it uses old media prefab information. This needs to be updated to the new options

AngeloBetty (approved):
Tested and it works

BettyBrian (approved):

PR #2129, opened 2022-07-20
feat: updated deleteRecord button usage

robbertruiter:


BettyBrian (changes requested):
I don't think the yarn.lock should be added. Can you revert the yarn.lock to it's original form?

BettyBrian (approved):

PR #2143, opened 2022-07-22
Drawer component mobile visibility

joerigroot:
When using the drawer component you can chose “Persistant” or “Temporary” mode and then you can chose a breakpoint on which you want the sidebar to be “Temporary”. When it is persistant you want it to be shown so the visibility is set to true. On mobile, when it’s temporary you need to have a separate possibility to set the default visibility. The most obvious choice is that it’s closed by default. 

I added an extra option “Toggle responsive visibility” to the component.

I also needed this feature in a use case (availability planning app). This app will be used a lot on mobile devices and it will also become a template.

### DIFF
--- a/src/components/drawer.js
+++ b/src/components/drawer.js
@@ -5,6 +5,7 @@
   orientation: 'HORIZONTAL',
   jsx: (() => {
     const { Children, env, useText } = B;
+    const { useMediaQuery, useTheme } = window.MaterialUI.Core;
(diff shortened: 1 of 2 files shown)


robbertruiter:
The changes have been made in drawer.js. Currently the component has been updated to TypeScript so this PR is not applicable anymore.
We'll take a look in implementing this fix later in our sprint

JorisPannekeet:

PR #2221, opened 2022-08-04
Feat/header and footer to TSX TMPLT-1538

robbertruiter:


src/prefabs/pageWithHeaderAndFooter.tsx
```diff
@@ -0,0 +1,530 @@
...
+  Text as TextPrefab,
+  textOptions,
+  AppBar,
+  appBarOptions,
+  OpenPageButton,
+  openPageButtonOptions,
+} from './structures';
+
+const attrs = {
+  name: 'Header and footer new',
```
sventruschel, line 37: The name of the template shouldnt change

src/prefabs/partials/footer.js
```diff
@@ -83,7 +80,7 @@
           type: 'SIZES',
         },
         {
-          value: ['M', 'M', 'M', 'M'],
+          value: ['L', 'L', 'L', 'L'],
```
SanderVanRiessen, line 83: I don't think this value should be on L, just leave this on size M

BettyBrian (approved):

PR #2230, opened 2022-08-08
Feat/login step pro coder api page 2105

jrquak:


src/prefabs/structures/ActionJSForm/index.ts
```diff
@@ -3,6 +3,12 @@ import { FormErrorAlert, FormSuccessAlert } from '../Alert';
 import { options } from './options';
 
 export const Form = (label: string) => {
+  if (label === 'Login Form Beta') {
```
benjamin-rocks, line 6: I think this is too implicit. I mean that is may not be clear to the user for this function that behavior changes based on the label you give the form.

We should have a specific flag to decide weather or not to show the success Alert.

```ts
export const Form = (label: string, showSuccessAlert: bool = true) => {
```

benjamin-rocks (approved):

src/prefabs/structures/ActionJSForm/index.ts
ingmar-stipriaan: I see that this is solved in the last commit from Jordy

benjamin-rocks (approved):

PR #2313, opened 2022-09-01
feat: crud with dialogs to tsx

BettyBrian:


bettybas (approved):
Both pages looked the same. Didn't find any errors.

PR #2327, opened 2022-09-05
feat: cusrom form without model

allamaley:


src/prefabs/form.tsx
```diff
@@ -63,206 +71,297 @@ const beforeCreate = ({
...
-            allowRelations
-            disabledKinds={[]}
-            disabledNames={['created_at', 'id', 'updated_at']}
-            modelId={modelId}
-            onChange={setProperties}
-            scopedModels={false}
-            value={properties}
-          />
+        <Field label="Model based form">
+          {/* TODO: fix onChange bug and move setmodelBased to Toggle onChange */}
```
stefan-betty, line 91: Do we need this TODO?

src/prefabs/form.tsx
```diff
@@ -63,206 +71,297 @@ const beforeCreate = ({
...
-            }
-            // eslint-disable-next-line no-console
-            return console.warn('PropertyKind not found');
-          });
-
+              // eslint-disable-next-line no-console
+              return console.warn('PropertyKind not found');
+            });
+          } else {
+            structure.descendants.push(cloneStructure(BettyPrefabs.STRING));
```
stefan-betty, line 363: What does this do?

src/prefabs/form.tsx
allamaley: yes, because this is a work around and doesnt suppose to work like this (onChange should be on Toogle)

src/prefabs/form.tsx
allamaley: it will add a Textfield component to the canvas, as a sibling right before Submit button (need this in acc criteria)

stefan-betty (approved):

src/prefabs/form.tsx
```diff
@@ -33,34 +39,12 @@ const beforeCreate = ({
...
-    'PRICE_EXPRESSION',
-    'RICH_TEXT',
-    'SERIAL',
-    'SIGNED_PDF',
-    'STRING_EXPRESSION',
-    'SUM',
-    'TEXT_EXPRESSION',
-    'ZIPCODE',
-  ];
-
```
benjamin-rocks, line 59: these should remain, we need them to exclude old types

src/prefabs/form.tsx
```diff
@@ -87,211 +71,308 @@ const beforeCreate = ({
...
+                onChange={(id) => {
+                  setModelId(id);
+                }}
+                margin
+              />
+            </Field>
+            <Field label="Select properties">
+              <PropertiesSelector
+                allowRelations
+                disabledKinds={[]}
```
benjamin-rocks, line 116: This is where we use them

benjamin-rocks (dismissed):

benjamin-rocks (approved):

PR #2349, opened 2022-09-07
chore: fixed compilation error

allamaley:


src/prefabs/form.tsx
```diff
@@ -63,11 +63,11 @@ const beforeCreate = ({
     'ZIPCODE',
   ];
 
-  const [modelId, setModelId] = React.useState(null);
-  const [model, setModel] = React.useState(null);
+  const [modelId, setModelId] = React.useState('');
+  const [model, setModel] = React.useState('');
```
stefan-betty, line 67: Shouldn't this be an empty object?
And why can't we keep it null?

benjamin-rocks (approved):

src/prefabs/form.tsx
allamaley: otherwise the compilation will fail, namely this mutation `validatePageArtifactMutation`

stefan-betty: We can also fix that in the backend, by allowing null.

PR #2411, opened 2022-09-20
Feat/user can use relational beta inputs page 1960

ingmar-stipriaan:


jrquak (approved):

src/components/selectInput.js
```diff
@@ -35,15 +36,20 @@
...
     const labelText = useText(label);
     const defaultValueText = useText(prefabValue);
     const helperTextResolved = useText(helperText);
     const validationMessageText = useText(validationValueMissing);
     const dataComponentAttributeValue = useText(dataComponentAttribute);
 
-    const { referenceModelId, modelId, kind, values = [] } = modelProperty;
+    const {
+      referenceModelId,
+      modelId = modelContext || '',
```
stefan-betty, line 49: Why do we assign modelContext or an empty string to modelId, why not modelId itself?

ingmar-stipriaan: Good point, we change this in the next MR

PR #2425, opened 2022-09-23
feat(PAGE-2238): Add permission & authProfId args to prepareAction fn.

stefan-betty:
The prepareAction helper function supports two new arguments: permissions and pageAuthenticationProfileId.
Both arguments are used to generate permissions. When permission is set to 'private' it will set the action to private and generate a permission for the admin role, so it can execute.
The `inherit` key is used to generate an admin role based on the page authentication profile. If there is a page authentication profile, it will create a execute permission for the admin role.

### DIFF
--- a/src/prefabs/createForm.tsx
+++ b/src/prefabs/createForm.tsx
@@ -38,6 +38,7 @@ const beforeCreate = ({
   const [model, setModel] = React.useState(null);
   const [idProperty, setIdProperty] = React.useState(null);
   const [properties, setProperties] = React.useState([]);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [permissions, setPermissions] = React.useState('private');
 
   const [validationMessage, setValidationMessage] = React.useState('');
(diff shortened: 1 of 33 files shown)


src/prefabs/createForm.tsx
```diff
@@ -37,10 +38,14 @@ const beforeCreate = ({
   const [model, setModel] = React.useState(null);
   const [idProperty, setIdProperty] = React.useState(null);
   const [properties, setProperties] = React.useState([]);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [permissions, setPermissions] = React.useState('private');
```
benjamin-rocks, line 42: which is the unused variable? maybe we can remove it.

stefan-betty: It is the setPermissions. While it is not used right now, this should give the indication that the permission is meant to be set by pro-coders.

benjamin-rocks (approved):

PR #2448, opened 2022-10-05
fix: show onerror interaction

allamaley:


src/components/fileUploadInput.js
```diff
@@ -76,8 +79,8 @@
         if (Array.isArray(error) && error.length === 0) {
           return;
         }
-        B.triggerEvent('onError', error);
-        setValidationMessage(`An error occured: ${error}`);
+        B.triggerEvent('onError');
```
benjamin-rocks, line 82: why not send the `error` as well?

stefan-betty (approved):

PR #2459, opened 2022-10-10
Fix/file input required is not set page 2217

ingmar-stipriaan:


src/components/actionJSButton.js
```diff
@@ -134,7 +129,7 @@
     };
 
     const buttonProps = {
-      disabled: disabled || isLoading,
+      disabled: disabled || isLoading || loading,
```
stefan-betty, line 132: You can use `showIndicator` here.

src/components/actionJSButton.js
```diff
@@ -124,7 +119,7 @@
       return undefined;
     };
 
-    const showIndicator = isLoading;
+    const showIndicator = isLoading || loading;
```
stefan-betty, line 122: Why do we have two loading variables?

ingmar-stipriaan: isLoading can be used by the builder to set te loading state. Loading will be returned from  useActionJS when the action is running, but doesn't set the state and useActionJS doesn't have an onLoading option to set te state either.

stefan-betty: I think we should change the name of `loading` to `isActionLoading` to make it clear what we are waiting for!

ingmar-stipriaan: Changed it to isLoadingAction

src/components/actionJSButton.js
ingmar-stipriaan: I used showIndicator now

jrquak (approved):

PR #2480, opened 2022-10-20
Feat/move input wizard to component set page 2210

allamaley:


src/prefabs/selectinput.tsx
```diff
@@ -1,32 +1,338 @@
...
+    Label,
+    TextInput: Text,
+    CircleQuestion,
+    BBTooltip,
+  },
   prefab: originalPrefab,
   save,
-}: BeforeCreateArgs) => {
+  helpers,
+}: any) => {
```
benjamin-rocks, line 30:
```
}: BeforeCreateArgs)
```

src/prefabs/selectinput.tsx
```diff
@@ -1,32 +1,338 @@
...
+              },
+            }));
+          }
+
+          if (
+            (originalPrefab.name === BettyPrefabs.AUTO_COMPLETE ||
+              originalPrefab.name === BettyPrefabs.SELECT ||
+              originalPrefab.name === BettyPrefabs.RADIO ||
+              originalPrefab.name === BettyPrefabs.MULTI_AUTO_COMPLETE) &&
+            propertyBased
```
benjamin-rocks, line 263: we don't need this code anymore since we know the prefab in question

src/prefabs/textinput.tsx
```diff
@@ -1,32 +1,334 @@
...
+    Field,
+    Footer,
+    Header,
+    FormField,
+    Toggle,
+    PropertySelector,
+    Label,
+    TextInput: Text,
+    CircleQuestion,
+    BBTooltip,
```
benjamin-rocks, line 25: is this a new export from the IDE? If so it should be documented.

benjamin-rocks (changes requested):

src/prefabs/textinput.tsx
```diff
@@ -2,31 +2,337 @@ import * as React from 'react';
...
+            originalPrefab.name === BettyPrefabs.RADIO
+          ) {
+            setOption(newPrefab.structure[0], 'property', (option) => ({
+              ...option,
+              value: {
+                id: propertyId,
+                type: DynamicTypeEnum.Property,
+                componentId: selectedPrefab?.id,
+              },
+            }));
```
benjamin-rocks, line 255: DynamicTypeEnum is a unnecessary here since we only use 'STRING' here.

Also I think we should probably introduce helpers to create options so that we can save the pro-coder from having to deal with these details.

```ts
setOption(newPrefab.structure[0], 'property', (option) => ({
              ...option,
              ...propertyOption(propertyId, selectedPrefab?.id)
            }));
```

src/prefabs/selectinput.tsx
```diff
@@ -2,31 +2,340 @@ import * as React from 'react';
 import { BeforeCreateArgs, Icon, prefab } from '@betty-blocks/component-sdk';
 import { SelectInput } from './structures/SelectInput';
 
+export enum DynamicTypeEnum {
+  Variable = 'VARIABLE',
+  Property = 'PROPERTY',
+  MeProperty = 'ME_PROPERTY',
+  Translation = 'TRANSLATION',
+}
+
```
benjamin-rocks, line 11: We should provide this in a helper prop as a constant. like we do with `BettyInput`.

allamaley: I got rid of `DynamicTypeEnum`, since indeed we use only **`"PROPERTY"`** option from there

src/prefabs/textinput.tsx
allamaley: I resolved first, but the second I think we can leave up to paradigm when they will extend it to other inputs

benjamin-rocks (approved):

PR #2501, opened 2022-10-31
fix: list item not showing Property name in list with data prefab

ThomasDekkerbetty:


src/prefabs/listWithData.tsx
```diff
@@ -21,9 +21,28 @@ const beforeCreate = ({
...
   close,
+  helpers: { useModelQuery },
 }: BeforeCreateArgs) => {
   const [modelId, setModelId] = React.useState('');
-  const [property, setProperty] = React.useState('');
+  const [primaryProperty, setPrimaryProperty] = React.useState('');
+
+  const { data } = useModelQuery({
+    variables: { id: modelId },
+    skip: !modelId,
```
robbertruiter, line 31: This line is not used. Why would you send a boolean value as a skip?

robbertruiter (approved):

PR #2565, opened 2022-11-16
Feat/multi select data table edge

vdberganton:


vdberganton:
To get this flow to work you need following:

- Data table with row selection enabled
- Action button with an array action input variable
- OnRowSelect > SetValue > Action Button interaction

This will send an array of the selected record Id’s to the action. There you currently still need a custom function to handle the deletion of those selected records but I’ll see if I can push that to the public store soon.

JorisPannekeet:

PR #2580, opened 2022-11-22
fix: setShowPagination (in case of "When needed") to false OR true

RobD-tech:
In case you have a data list of 5 records and the pagination component option is set to "When needed" and the rows per page is equal to 5, no pagination is shown. When you add a new (6th) record via a form/action button and trigger a refetch on the data list, the pagination will now show up since you'll have two pages to show. On the contrary, when you have 6 records and delete one, the pagination will disappear.


robbertruiter (approved):
Looks and works good!

PR #2582, opened 2022-11-22
fix: setShowPagination when needed false OR true

RobD-tech:
Conditionally show or hide pagination "when needed" after refetch of the data list.


robbertruiter (approved):
Works great!

PR #2679, opened 2023-01-05
fix: add tmp folder to gitignore TMPLT-2012

robbertruiter:


SanderVanRiessen (changes requested):
@ewoutquax 
Do we really need / want this? Because for every build the component set is saving a build in the tmp folder. How does this work in the production, acceptance, and edge environment? Are all builds still saved in the tmp folder when someone does a merge to one of the environments?

SanderVanRiessen (approved):

PR #2757, opened 2023-02-07
Feat/add label number of pages option

RobD-tech:
The Data Table component already has the option to set the label of the pagination via a variable component option, but the Data List has not. This option is useful while creating a Dutch or multi-language application because "2-4 of 8" would look odd.


robbertruiter (approved):
Looks amazing

PR #2767, opened 2023-02-09
feat: Add property whitelist to filter

chrisobdam:


sventruschel (changes requested):
The option is not added to the prefab

sventruschel (approved):

PR #2782, opened 2023-02-15
Feat add themeweight to text

chrisobdam:


SanderVanRiessen:
I created a new PR to add the themeweight to the text component, because there are a lot of changes in this PR that are already merged to production.
This is the new PR: https://github.com/bettyblocks/material-ui-component-set/pull/2837

PR #2796, opened 2023-02-21
Feat/on no results data listimprovement

robinvanderburght:
In the current situation when using the has-many relation option doesn't triggers the onSucces and onNoResults interactions. This change does triggers them with the regular data fetch option, as well as using the has-many relation option.

### DIFF
--- a/src/components/dataList.js
+++ b/src/components/dataList.js
@@ -268,7 +268,6 @@
             variables: {
               ...(orderByPath ? { sort: { relation: sort } } : {}),
             },
-            onCompleted() {},
             onError(resp) {
               if (!displayError) {
                 B.triggerEvent('onError', resp);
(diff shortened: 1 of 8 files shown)


src/components/dataList.js
```diff
@@ -268,14 +268,7 @@
...
             },
-            onCompleted(res) {
-              const hasResult = res && res.results && res.results.length > 0;
-              if (hasResult) {
-                B.triggerEvent('onSuccess', res.results);
-              } else {
-                B.triggerEvent('onNoResults');
-              }
-            },
+            onCompleted() {},
```
sventruschel, line 271: Can't this be removed now?  doesnt look like it does anything anymore

sventruschel (approved):

PR #2809, opened 2023-02-28
Feat/record action input update form page 2593

ingmar-stipriaan:


robbertruiter (commented):
It doesn't cover all the page templates that have an update form

jrquak (approved):

jrquak (approved):

PR #2869, opened 2023-03-20
Edge test components

ingmar-stipriaan:


src/prefabs/structures/AllOptions/options/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    ],
+    {
+      value: 'start',
+      configuration: {
+        condition: hideIf('icon', 'EQ', 'None'),
+        showOnDrop: true,
+      },
+    },
+  ),
+  // actionVariableId: option('ACTION_JS_VARIABLE', {
```
stefan-betty, line 90: What is the idea of this?

stefan-betty (dismissed):

PR #2917, opened 2023-03-30
chore: add multiple choice question

jrquak:


src/components/radioinput.js
```diff
@@ -36,7 +37,8 @@
       RadioGroup,
     } = window.MaterialUI.Core;
     const isDev = env === 'dev';
-    const modelProperty = getProperty(actionProperty.modelProperty || '') || {};
```
robbertruiter, line 39: Why still support the `actionProperty` when that won't be sent to the component anymore since you changed it to `property`?

robbertruiter (changes requested):

PR #2998, opened 2023-05-09
Feat/questionnaire

Rahmi599:


src/prefabs/structures/TextInput/options/styles.ts
```diff
@@ -91,7 +91,7 @@ export const styles = {
   }),
 
   placeholderColor: color('Placeholder color', {
-    value: ThemeColor.LIGHT,
+    value: ThemeColor.BLACK,
```
robbertruiter, line 94: Where do we specify that we want only the text input to always have a black label instead of a light one?
Is this a request by UX that is not in the ticket?
If so, this should be changed in a branch that goes further than only edge.
If it should only be changed in the questionnaire template, please update the option over there instead of the global one

robbertruiter (changes requested):

PR #3002, opened 2023-05-09
fix: remove duplicate icons

BettyBrian:


.github/workflows/publish_bundle_acceptance.yml
```diff
@@ -0,0 +1,28 @@
...
+      node-version: 16.11.1
+      - name: build
+      run: yarn build
+
+    - name: bundle
+    run: yarn bundle
+
+    - name: upload
+    env:
+      AZURE_BLOB_ACCOUNT: ${{ secrets.AZURE_BLOB_ACCOUNT_EDGE }}
```
robbertruiter, line 26: Is it intentional to use the Azure blob account secret for edge while being in the pulbish_budnle_acceptance?

robbertruiter (changes requested):

SanderVanRiessen (changes requested):

SanderVanRiessen (changes requested):
The CLI version in the package json needs to be up to date.

SanderVanRiessen (approved):

PR #3068, opened 2023-06-05
fix: tabs responsive font

rutgerlangedijk:


sventruschel:
Closed because this already seems fixed

PR #3100, opened 2023-06-20
fix: autcomplete is using old value

SanderVanRiessen:


jrquak (approved):

ingmar-stipriaan (approved):

src/components/autocompleteInput.js
```diff
@@ -579,23 +580,23 @@
     /*
      * Convert `value` state into something the `value` prop of the `Autocomplete` component will accept with the right settings
      */
-    const getValue = () => {
+    const getValue = (testVAlue = value) => {
       if (isListProperty) {
-        return value;
+        return testVAlue;
```
jrquak, line 585: small typo

jrquak: + maybe rename this to a better variable name

src/components/autocompleteInput.js
```diff
@@ -622,6 +623,36 @@
...
+
+      if (!isListProperty) {
+        if (currentValue !== debouncedCurrentValue) {
+          debounceCurrentValue = setTimeout(() => {
+            setDebouncedCurrentValue(currentValue);
+          }, 250);
+        } else {
+          let triggerEventValue;
+
+          if (!isListProperty) {
```
stefan-betty, line 637: you already checked what `!isListProperty` is, right?

src/components/autocompleteInput.js
```diff
@@ -622,6 +623,36 @@
...
+            triggerEventValue = debounceCurrentValue
+              ? debounceCurrentValue[valueProp.name]
+              : '';
+          } else if (isListProperty) {
```
stefan-betty, line 641: this will never be true

src/components/autocompleteInput.js
SanderVanRiessen: This should be fixed in the next commit

src/components/autocompleteInput.js
stefan-betty, line 641: We can never enter this else-if statement.
If `isListProperty` is false, the conditional-check on line 629 is valid. 
And now we are checking: 
`else if(false)` which ofcourse does not work.

stefan-betty: So this can be removed?

src/components/autocompleteInput.js
```diff
@@ -622,6 +623,36 @@
...
+      if (!isListProperty) {
+        if (currentValue !== debouncedCurrentValue) {
+          debounceCurrentValue = setTimeout(() => {
+            setDebouncedCurrentValue(currentValue);
+          }, 250);
+        } else {
+          let triggerEventValue;
+
+          if (!isListProperty) {
+            triggerEventValue = debounceCurrentValue
```
stefan-betty, line 638: Isn't `debounceCurrentValue` always undefined since you define it on line 627?

src/components/autocompleteInput.js
SanderVanRiessen: I have removed the first check since we want to use the onchange interaction with the list property and relation property

src/components/autocompleteInput.js
SanderVanRiessen: I have removed the first check since we want to use the onchange interaction with the list property and relation property

PR #3104, opened 2023-06-21
fix: option label relation in select

SanderVanRiessen:


src/components/selectInput.js
```diff
@@ -255,6 +255,38 @@
...
+    const idOrPathLabel =
+      typeof labelProperty.id !== 'undefined'
+        ? labelProperty.id
+        : labelProperty;
+    const labelPropertyPath =
+      typeof idOrPathLabel === 'string' ? [idOrPathLabel] : idOrPathLabel;
+
+    const renderLabel = (option) => {
+      let optionLabel = '';
+      if (labelPropertyPath.length === 1 && labelPropertyPath[0] !== '') {
```
jrquak, line 267: Is it possible to move this check to a const?

jrquak (approved):

PR #3107, opened 2023-06-22
fix: return fallback if acc is null

SanderVanRiessen:


src/components/autocompleteInput.js
```diff
@@ -679,6 +679,9 @@
         labelPropertyPath.length > 0 && labelPropertyPath[0] !== '';
       if (emptyPropertyPath) {
         optionLabel = labelPropertyPath.reduce((acc, propertyId) => {
+          if (!acc) {
```
stefan-betty, line 682: You define `acc` as `option` (line 686). Is it possible that `option` can be `null`/?

stefan-betty (approved):

PR #3134, opened 2023-07-04
Release

dylanBetty:


SanderVanRiessen (dismissed):

dylanBetty:
Double checking if this maybe the sync can be fixed via the `usePageState` hook rather than in the component

PR #3136, opened 2023-07-04
Fix for the horizontal scrollbars

robinvanderburght:
Currently when using the margin none option and dragging the select in a column the column get's horizontal scrollbars because of the hidden input where the value is set is wider then the select itself. This change fixes this issue.


RobD-tech:
Duplicate from https://github.com/bettyblocks/material-ui-component-set/pull/3197

PR #3140, opened 2023-07-07
feat: Store datetime in ISO-format

ewoutquax:
The ISO-format contains timezone information.
This will solve a bug where entering a datetime in a browser in another timezone as the server, would result in a timejump when requesting the datetime.


dylanBetty (approved):

dylanBetty:
Approved, although It won't fix existing components, but just new ones

PR #3141, opened 2023-07-07
feat: remove betty-jwt cookie, when logging in via user/password

ewoutquax:
Allowing to login via only one profile at a time, will prevent confusion in the /me-query. This query would often use the wrong keys, when multiple keys were present, and show unexpected behaviour.

### DIFF
--- a/src/interactions/login.ts
+++ b/src/interactions/login.ts
@@ -34,7 +34,7 @@ function login({
 
     const d = new Date();
     const e = d.toUTCString();
-    document.cookie = `betty_jwt=dummy; expires=${e}; SameSite=None`;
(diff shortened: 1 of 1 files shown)


src/interactions/login.ts
```diff
@@ -32,6 +32,10 @@ function login({
       document.cookie = `BBLocale=${decodedToken.locale};path=/`;
     }
 
+    const d = new Date();
+    const e = d.toUTCString();
+    document.cookie = `betty_jwt=dummy; expires=${e}; SameSite=None`;
```
dylanBetty, line 37: Let's remove betty_jwt, or rename `dummy` to something like `None`

dylanBetty (approved):

PR #3150, opened 2023-07-13
chore: add fileupload to form reconfigure

ingmar-stipriaan:


jrquak (approved):

src/prefabs/structures/FileUpload/index.ts
```diff
@@ -1,59 +1,41 @@
...
-  if (config.supportImages) {
-    options.accept = updateOption(options.accept, {
-      value: ['image/*'],
-    });
-  }
+export const FileUpload = (
+  config: Configuration = {},
+  descendants: PrefabReference[] = [],
+) => {
+  const options = { ...(config.options || fileUploadOptions) };
```
stefan-betty, line 33: Shouldn't we apply config.options to the defaultOptions? So that we overwrite?

```javascript
{...fileUploadOptions, ...config.options}
````

File: src/prefabs/imageUploadInput.tsx [hunk 2]
```diff
@@ -226,6 +235,12 @@ const beforeCreate = ({
               },
             }),
           );
+
+          setOption(newPrefab.structure[0], 'accept', (option) => ({
+            ...option,
```
stefan-betty, line 240: What are we doing here? why is it needed?

PR #3205, opened 2023-08-01
feat: add option to show text when no results fetched PFR-697

RobD-tech:
Currently, there is no option in a Data List or Data table PB component to show a (default) message when there are no results fetched from the Data API. The only way to properly show that there are no results, is to use an “OnNoResults” interaction in combination with a Conditional component and a Text component. By adding a variable option to these two Data components, builders can save an interaction and (at least) two additional components for each Data List/Table which can help reduce the page complexity.

### DIFF
--- a/src/components/dataTable.js
+++ b/src/components/dataTable.js
@@ -500,7 +500,7 @@
 
       if (results.length === 0 && parsedNoResultsText) {
         return (
-          <TableRow classes={{ root: classes.bodyRow }}>
+          <TableRow>
             <TableCell>{parsedNoResultsText}</TableCell>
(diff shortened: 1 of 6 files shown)


src/components/dataTable.js
```diff
@@ -496,6 +498,14 @@
         ));
       }
 
+      if (results.length === 0 && parsedNoResultsText) {
+        return (
+          <TableRow classes={{ root: classes.bodyRow }}>
```
stefan-betty, line 503: Why do we need to overwrite the style here?

jrquak (approved):

PR #3208, opened 2023-08-01
fix: overflow option added for allowing sticky

Reinier1982:
Added an extra option to the column advanced options. When deselecting this option, it removes the CSS class 'overflow', so you can use 'display sticky' in your box, so you can have menu's that scroll with your content.


SanderVanRiessen:
I'm closing this PR because according the UX it's not the right way to add this as a checkbox option to the column.

PR #3252, opened 2023-10-11
fix: properly display error msg in data table header PFR-727

RobD-tech:
This fix makes sure the error message is entirely shown in the header of the data table component. An example of the current (before) and new (after) situation can be seen in [PFR-727](https://bettyblocks.atlassian.net/browse/PFR-727).

### DIFF
--- a/src/components/dataTable.js
+++ b/src/components/dataTable.js
@@ -469,15 +469,22 @@
     };
 
     const renderTableHead = () => {
-      if ((loading && !loadOnScroll) || error) {
+      if (loading && !loadOnScroll) {
         return Array.from(Array(children.length).keys()).map((colIdx) => (
           <TableCell key={colIdx}>
-            <div className={classes.skeleton}>
-              {error && displayError && error.message}
-            </div>
+            <div className={classes.skeleton} />
           </TableCell>
         ));
       }
+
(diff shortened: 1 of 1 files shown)


src/components/dataTable.js
```diff
@@ -469,15 +469,22 @@
...
-              {error && displayError && error.message}
-            </div>
+            <div className={classes.skeleton} />
           </TableCell>
         ));
       }
+
+      if (error) {
+        return (
+          <TableCell colSpan={children.length}>
```
M-Ryan92, line 482: you are missing a key prop here like how its done above.
`<TableCell key={colIdx}>`

src/components/dataTable.js
```diff
@@ -469,15 +469,22 @@
...
+            {error && displayError && error.message}
```
M-Ryan92, line 483: when displayError is falsy what should happen then? for the user it might be confusing if the cell stays empty. but im not sure about that.

src/components/dataTable.js
MenouerBetty: That is only needed for mapped components from a array @M-Ryan92

JorisPannekeet (approved):

PR #3259, opened 2023-10-16
Z-index option on Box component

robinvanderburght:
This feature adds a z-index option when the 'fixed' Position option is selected

### DIFF
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -141,6 +141,8 @@
         minHeight: 0,
         position: ({ options: { position } }) =>
           (!isDev && position) || 'unset',
+        zIndex: ({ options: { zIndex, position } }) =>
+          position === 'fixed' ? zIndex : 0,
         top: ({ options: { top } }) => !isDev && top,
         right: ({ options: { right } }) => !isDev && right,
         bottom: ({ options: { bottom } }) => !isDev && bottom,
--- a/src/components/box.js
+++ b/src/components/box.js
@@ -141,8 +141,12 @@
         minHeight: 0,
         position: ({ options: { position } }) =>
           (!isDev && position) || 'unset',
-        zIndex: ({ options: { zIndex, position } }) =>
-          position === 'fixed' ? zIndex : 0,
+        zIndex: ({ options: { zIndex, position } }) => {
+          if (zIndex === '') {
+            return 'auto';
+          }
+          return position !== 'static' ? zIndex : 'auto';
+        },
         top: ({ options: { top } }) => !isDev && top,
         right: ({ options: { right } }) => !isDev && right,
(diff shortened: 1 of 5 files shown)


src/prefabs/structures/Box/options/index.ts
```diff
@@ -145,6 +147,13 @@ export const boxOptions = {
       },
     },
   ),
+  zIndex: number('Z-index', {
+    value: '0',
+    configuration: {
+      condition: showIf('position', 'EQ', 'fixed'),
```
dylanBetty, line 153: I think this condition should be this:
```typescript
condition: showIf('position', 'NEQ', 'static'),
```
You want to always show the zIndex, except if the position is set to 'static'

src/components/box.js
```diff
@@ -141,6 +141,8 @@
         minHeight: 0,
         position: ({ options: { position } }) =>
           (!isDev && position) || 'unset',
+        zIndex: ({ options: { zIndex, position } }) =>
+          position === 'fixed' ? zIndex : 0,
```
dylanBetty, line 145: The zIndex value should only be applied if a position (other than `static`) is used. I think this should be:
```javascript
position !== 'static' ? zIndex : 'auto'
```

src/prefabs/structures/Box/options/index.ts
```diff
@@ -145,6 +147,13 @@ export const boxOptions = {
       },
     },
   ),
+  zIndex: number('Z-index', {
+    value: '0',
```
dylanBetty, line 151: The default z-Index should be set to `auto` by default: https://developer.mozilla.org/en-US/docs/Web/CSS/z-index
But because this is a number input, `auto` won't be a valid value.
I'm thinking it might be nice to set this to an empty string by default, and then (in src/components/box) add an extra if statement:
```javascript
if (zIndex === '') {
  return auto;
}
```
This way we can still make use of a number option, but still fall back to 'auto' if no value is filled in :)

src/prefabs/structures/Box/options/index.ts
robinvanderburght: @dylanBetty appearently NEQ doesn't exists, therefore I choose the option to use 
`condition: hideIf('position', 'EQ', 'static')`

src/prefabs/structures/Box/options/index.ts
```diff
@@ -2,13 +2,16 @@ import {
...
   color,
+  number,
   option,
   ThemeColor,
   toggle,
   variable,
   buttongroup,
   dropdown,
   displayLogic,
+  showIf,
```
dylanBetty, line 13: This can probably be removed

dylanBetty (approved):

PR #3266, opened 2023-10-25
Fix/improve visibility delete icon in chip autocomplete pfr 737

RobD-tech:
When the chips in the autocomplete have a dark background color, the delete icon is not (clearly) visible. 

If the background color is dark, you will apply a light color to the text within the chip and the icon should behave the same. By default, the background color of the chip is set to Light (from the theme builder). The default Material UI color for the delete icon is fine then, but as soon as the background color of the chip is changed the icon color should be the same as the text color. This was a finding when working with the design system from the police. The UI was agreed upon with UX.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# [2.165.0](https://github.com/bettyblocks/material-ui-component-set/compare/v2.164.0...v2.165.0) (2023-10-17)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


stefan-betty:
Before we merge this to edge, could you merge acceptance into your branch?

stefan-betty (approved):

ingmar-stipriaan (dismissed):

PR #3312, opened 2024-01-02
Feat/add data list and data table selected record component states page 3992

dylanBetty:


src/prefabs/hiddeninput.tsx
JorisPannekeet, line 1: Why changes to hiddenInput prefab?

JorisPannekeet (approved):

src/prefabs/hiddeninput.tsx
dylanBetty: These are from a different commit that has not been downstreamed yet

PR #3350, opened 2024-01-29
fix: show correct date in picker input

jrquak:


src/components/dateTimePickerInput.js
```diff
@@ -117,7 +117,12 @@
         switch (type) {
           case 'date': {
             const formattedDate = DateFns.parse(parsedValue, dateFormat);
-            setSelectedDate(formattedDate);
+
+            if (isValidDate(formattedDate)) {
+              setSelectedDate(formattedDate);
+            } else {
+              setSelectedDate(new Date(parsedValue.replace(/-/g, '/')));
```
ewoutquax, line 124: Is it a valid solution to move this line to 119? And then remove the `isValidDate` check?

ingmar-stipriaan (approved):

PR #3356, opened 2024-01-31
Feat/filter pagevariable properties by kind page 3764

ingmar-stipriaan:


src/prefabs/checkboxGroupInput.tsx
```diff
@@ -349,5 +349,15 @@ const attributes = {
...
 export default prefab('Checkbox Group', attributes, beforeCreate, [
-  CheckboxGroup({ label: 'Checkbox Group' }),
+  CheckboxGroup({
+    label: 'Checkbox Group',
+    value: {
+      value: [],
+      configuration: {
+        allowRelations: true,
+        allowedClickThroughKinds: ['HAS_MANY', 'HAS_AND_BELONGS_TO_MANY'],
+        allowedSplitButtonKinds: [],
```
stefan-betty, line 359: What does an empty array mean? disallow everything?

src/prefabs/structures/CheckboxGroup/index.ts
```diff
@@ -37,6 +39,10 @@ export const CheckboxGroup = (
     },
   ];
 
+  if (config.value) {
+    options.value = updateOption(options.value, { ...config.value });
```
stefan-betty, line 43: We do we set the config value into options?

src/prefabs/checkboxGroupInput.tsx
ingmar-stipriaan: yes

src/prefabs/structures/CheckboxGroup/index.ts
ingmar-stipriaan: so that you can reuse the component for other prefabs and overwrite the default value and its settings like the allowedKinds

jrquak (approved):

PR #3401, opened 2024-03-22
Feat/use unique id to have valid label for attributes page 4088

dylanBetty:


src/components/textinput.js
```diff
@@ -268,6 +269,7 @@
       >
         {labelText && !hideLabel && (
           <InputLabel
+            htmlFor={id}
```
ewoutquax, line 272: ComponentID is shown in the snapshot as `secret-id-do-not-use`, but we do use it here.

I do not know exactly how valid the 'do not use'-warning is, but maybe we should take it seriously.

Alternative: make a page-wide counter-helper, which return a unique, incremented counter each time it is called. We can use the result of that counter, with a prefix, to generate unique `for`-ids.

Pro: this setup may also help for the iterated input-components.

PR #3410, opened 2024-04-29
Fix/TECHSUP-9010

degrootsam:
SEE: https://bettyblocks.atlassian.net/browse/TECHSUP-9010

TLDR
Options were not refetched after clearing the value. 

Cause:
The useEffect hook on line 666 is not triggered when inputValue === '' and thus the debouncedInputValue will not be set properly.

### DIFF
--- a/src/components/autocompleteInput.js
+++ b/src/components/autocompleteInput.js
@@ -97,7 +97,6 @@
     const [interactionFilter, setInteractionFilter] = useState({});
     const defaultValueEvaluatedRef = useRef(false);
     const { current: labelControlRef } = useRef(generateUUID());
-
     const isNumberType = type === 'number';
 
     const validPattern = pattern || null;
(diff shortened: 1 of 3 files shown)


degrootsam:
Apr-29-2024 14-40-13

src/components/autocompleteInput.js
```diff
@@ -756,13 +751,14 @@
...
             handleValidation(validation);
-            if (
-              event &&
-              (event.type === 'change' || event.type === 'keydown')
-            ) {
-              setInputValue(newValue);
-            } else if (event && event.type === 'click') {
-              setInputValue(newValue);
+            setInputValue(newValue || '');
+            // START TECHSUP-9010
```
jrquak, line 755: Looks good, but can you remove the comments?

jrquak (approved):

PR #3419, opened 2024-05-24
Feat: min and max date validation date picker PFR-824

RobD-tech:
This feature includes two new variable component options: "Minimum value" and "Maximum value". The values of these component options have to be formatted according to the component option "Format". In React, the component properties "minDate" and "maxDate" had to be added conditionally in order to get it to work properly, because if the value "null" is being passed to the component, i.e. `<DateTimeComponent minDate={null} maxDate={null}>...</DateTimeComponent>`, the navigation buttons will be disabled.

In the future it would be nice to have these validation options available for the datetime and time pickers as well, but the current version of these pickers unfortunately doesn't allow to incorporate the minimum and maximum validations.

### DIFF
--- a/src/components/dateTimePickerInput.js
+++ b/src/components/dateTimePickerInput.js
@@ -22,6 +22,8 @@
       size,
       fullWidth,
       required,
+      minValue,
+      maxValue,
       margin,
       helperText = [''],
       disableToolbar,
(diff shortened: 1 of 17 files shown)


src/components/dateTimePickerInput.js
```diff
@@ -172,11 +176,16 @@
...
     let resultString;
     let use24HourClock = true;
+    const validation = {};
 
     switch (type) {
       case 'date': {
         DateTimeComponent = KeyboardDatePicker;
         format = dateFormat || 'dd/MM/yyyy';
+        if (minValueText)
+          validation.minDate = DateFns.parse(minValueText, dateFormat);
```
jrquak, line 186: use let instead of const when modifying vars

RobD-tech: This is not a variable (re)assignment but a key-value pair addition to an object instead. How should I use "let" here?

jrquak: I looked at this before my Monday morning coffee :stuck_out_tongue: Approved

ingmar-stipriaan (approved):

PR #3426, opened 2024-06-19
fix: manually trigger updateIndicator to show active tab on page load TECHSUP-8534

RobD-tech:
This fix makes sure the active tab indicator is shown on page load. For more information, see [TECHSUP-8534](https://bettyblocks.atlassian.net/browse/TECHSUP-8534).

### DIFF
--- a/src/components/tabs.js
+++ b/src/components/tabs.js
@@ -48,7 +48,6 @@
     }, [isDev, currentTab]);
 
     useEffect(() => {
-      // Manually trigger updateIndicator() to show active tab on page load (TECHSUP-8534)
       setTimeout(() => {
         tabsRef.current.updateIndicator();
       }, 100);
--- a/src/components/tabs.js
+++ b/src/components/tabs.js
@@ -24,6 +24,7 @@
     const orientation =
       alignment === 'top' || alignment === 'bottom' ? 'horizontal' : 'vertical';
     const isDev = env === 'dev';
+    const tabsRef = useRef();
     const currentTab = isDev ? selectedDesignTabIndex : defaultValue;
     const [value, setValue] = useState(parseInt(currentTab - 1, 10) || 0);
     const [tabData, setTabData] = useState({});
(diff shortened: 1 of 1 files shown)


ingmar-stipriaan (approved):

src/components/tabs.js
```diff
@@ -46,6 +47,13 @@
       }
     }, [isDev, currentTab]);
 
+    useEffect(() => {
+      // Manually trigger updateIndicator() to show active tab on page load (TECHSUP-8534)
```
JorisPannekeet, line 51: Remove comment

src/components/tabs.js
```diff
@@ -46,6 +47,13 @@
...
+      setTimeout(() => {
```
JorisPannekeet, line 52: Why the timeout? wont a check like` if(tabsRef.current)` suffice?

src/components/tabs.js
RobD-tech: This comment was suggested by your team mate so that we can find it for future references, but I'm fine with removing it.

src/components/tabs.js
RobD-tech: @JorisPannekeet Please see the research I conducted in [TECHSUP-8534](https://bettyblocks.atlassian.net/browse/TECHSUP-8534?focusedCommentId=205814). Other implementations simply won't work.

ingmar-stipriaan (approved):

src/components/tabs.js
JorisPannekeet: @RobD-tech I see, in your research however i do not see the following implementation:
`useEffect(()=>{
 if(tabsRef.current){ // Check if the element is rendered
  // do logic
}
},[tabsRef.current]); 
`

Can you validate if that works? If not then the timeout is fine.

RobD-tech: @JorisPannekeet Unfortunately your implementation also doesn't work.

JorisPannekeet (approved):

src/components/tabs.js
JorisPannekeet: Alright, thanks for checking. I've approved the mr.

jeroenreijs: Why timeout 100? Why not 200? Or 50? Unless you have an explanation for that, it's not wise to use a solution like this.

RobD-tech: @jeroenreijs I have tested with various values and the fix becomes consistent with a value of 100 in the use cases that I've tested. You don't want this value to be too large because you don't want the end user to notice the delay, i.e. see the indicator appearing some time after the page has already loaded. The underlying reason why a timeout is needed to fix this bug is because the active tab indicator includes an animated transition which apparently influences it's appearance.

PR #3442, opened 2024-08-26
feat: set focus to text input component PFR-899

RobD-tech:
This feature includes a new checkbox component option to set auto focus on an input on page load, and an interaction called `Focus` to focus on an input field. When the component option is set to _true_ for multiple inputs, only the last input will be focused on.

`const focusHandler = () => setTimeout(() => { inputRef.current.focus(); }, 50);`
A short timeout is required to be able to trigger the focus interaction on inputs that are inside Dialog or DrawerSidebar components.

Link to the PFR ticket: [PFR-899](https://bettyblocks.atlassian.net/browse/PFR-899).

### DIFF
--- a/src/components/textinput.js
+++ b/src/components/textinput.js
@@ -209,7 +209,7 @@
     const focusHandler = () =>
       setTimeout(() => {
         inputRef.current.focus();
-      }, 50);
+      }, 0);
 
     B.defineFunction('Clear', () => setCurrentValue(''));
     B.defineFunction('Enable', () => setIsDisabled(false));
--- a/src/components/textinput.js
+++ b/src/components/textinput.js
@@ -7,6 +7,7 @@
     const {
       actionVariableId: name,
       autoComplete,
+      autoFocus,
       disabled,
       error,
       label,
@@ -67,6 +68,7 @@
     const parsedLabel = useText(label);
     const labelText = parsedLabel;
     const debouncedOnChangeRef = useRef(null);
+    const inputRef = useRef();
 
     const { current: labelControlRef } = useRef(generateUUID());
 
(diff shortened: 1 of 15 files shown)


src/components/textinput.js
```diff
@@ -204,10 +206,16 @@
       handleValidation(validity);
     };
 
+    const focusHandler = () =>
+      setTimeout(() => {
+        inputRef.current.focus();
+      }, 50);
```
stefan-betty, line 212: Perhaps its better to use `0` instead `50`.
See https://developer.mozilla.org/en-US/docs/Web/API/setTimeout
Or do you have a specific reason to use `50`?

RobD-tech: Sweet, didn't know about that. A timeout of `0` seemed to work consistently in my test cases as well. Added a new commit to this PR.

src/prefabs/structures/TextInput/options/validation.ts
```diff
@@ -35,6 +35,7 @@ export const validation = {
   }),
 
   autoComplete: toggle('Autocomplete', { value: false }),
+  autoFocus: toggle('Auto focus on input', { value: false }),
```
stefan-betty, line 38: Could you change the message to `Autofocus`?

stefan-betty (dismissed):

ingmar-stipriaan (approved):

stefan-betty (dismissed):

ingmar-stipriaan (approved):

stefan-betty (dismissed):

SH0PK33P3R (approved):

dylanBetty (dismissed):

ingmar-stipriaan (dismissed):

PR #3443, opened 2024-08-26
feat: month picker component PFR-643

RobD-tech:
A new prefab called **Month picker** which is based on the Date picker component and meant for a user to select an available Month only. Material UI offers the [“views” prop](https://mui.com/x/react-date-pickers/date-picker/#views) that makes it easy to configure the selection views in the pop up (which is one of “date”, “year”, “month”, “day” “hours”, “minutes”, “seconds” depending on the type of the picker).

In this PR the "views" array is provided for each prefab type so that it can easily be (re)configured in the future. For example, you currently can’t select the seconds by default with the Time picker because it is not included in the views array by default, but the seconds will be shown in the input because of the default format option value “HH:mm:ss“. You could either remove ":ss" from the default value of the format option or append "seconds" to the array of views to fix this in the future.

Link to the PFR ticket: [PFR-643](https://bettyblocks.atlassian.net/browse/PFR-643).

### DIFF
--- a/src/components/dateTimePickerInput.js
+++ b/src/components/dateTimePickerInput.js
@@ -19,6 +19,7 @@
       dateFormat,
       timeFormat,
       datetimeFormat,
+      monthFormat,
       size,
       fullWidth,
       required,
@@ -204,6 +205,19 @@
             break;
           }
 
+          case 'month': {
+            const formattedDate = DateFns.parse(parsedValue, monthFormat);
+
+            if (isValidDate(formattedDate)) {
+              setSelectedDate(formattedDate);
+            } else {
+              // convert to slashes because it conflicts with the MUI DateTimeCmp
+              const parsedValueWithSlashes = parsedValue.replace(/-/g, '/');
+              setSelectedDate(new Date(parsedValueWithSlashes));
+            }
+            break;
+          }
+
           default:
         }
       } else {
@@ -221,11 +235,13 @@
     let use24HourClock = true;
     let minDate;
     let maxDate;
+    let views;
 
     switch (type) {
       case 'date': {
         DateTimeComponent = KeyboardDatePicker;
         format = dateFormat || 'dd/MM/yyyy';
+        views = ['year', 'date'];
 
         minDate = convertToValidDate(minValueText);
         maxDate = convertToValidDate(maxValueText);
@@ -239,6 +255,7 @@
         DateTimeComponent = KeyboardDateTimePicker;
         format = datetimeFormat || 'dd/MM/yyyy HH:mm:ss';
         use24HourClock = use24HourClockTime;
+        views = ['year', 'date', 'hours', 'minutes'];
 
         resultString = isValidDate(selectedDate)
           ? selectedDate.toISOString()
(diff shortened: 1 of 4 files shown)


ingmar-stipriaan (approved):

ingmar-stipriaan (approved):

NickRuigrok (approved):

src/components/dateTimePickerInput.js
```diff
@@ -249,12 +266,23 @@
...
           ? DateFns.format(selectedDate, 'HH:mm:ss')
           : null;
         break;
       }
+      case 'month': {
+        DateTimeComponent = KeyboardDatePicker;
+        format = monthFormat || 'MMMM';
+        views = ['month'];
+
+        resultString = isValidDate(selectedDate)
```
stefan-betty, line 281: How can it be called `resultString` if a value that can be assigned is `null` ?:) 
If we have to assign `null`, and not an empty string `''`, then I would suggest to change the name to `result` or something like that!

src/prefabs/monthPickerInput.tsx
```diff
@@ -0,0 +1,338 @@
...
+    createBlacklist,
+    useModelRelationQuery,
+  } = helpers;
+
+  const [propertyPath, setProperty] = React.useState<any>('');
+  const [variableInput, setVariableInput] = React.useState(null);
+  const modelId = useModelIdSelector();
+  const actionId = useActionIdSelector();
+  const selectedPrefab = usePrefabSelector();
+  const [model, setModel] = React.useState<any>(null);
```
stefan-betty, line 48: I hope we know the type here? `any` could give us headaches later on! There is probably a `Model type` somewhere!

src/prefabs/monthPickerInput.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+    return true;
+  };
+
+  let name: string | undefined;
+  let propertyKind;
+  let propertyModelId;
+  const componentId = createUuid();
+
+  function isProperty(path: string) {
```
stefan-betty, line 77: This is conflicting with line 79. 
You state that the `path`  can only be of type `string`, and then at line 79 you check if its `not` a string.
I think you want something like:
`isProperty(path: Property | Array<string> | string`

src/prefabs/monthPickerInput.tsx
```diff
@@ -0,0 +1,338 @@
...
+        'Model details are still loading, please try submitting again.',
+      );
+      return false;
+    }
+
+    return true;
+  };
+
+  let name: string | undefined;
+  let propertyKind;
```
stefan-betty, line 73: Can we add types here as well? :)

src/prefabs/monthPickerInput.tsx
```diff
@@ -0,0 +1,338 @@
...
+  function isProperty(path: string) {
+    return (
+      typeof path !== 'string' &&
+      typeof path === 'object' &&
+      !Array.isArray(path)
+    );
+  }
+
+  let propertyId: string;
+  if (isProperty(propertyPath)) {
```
stefan-betty, line 86: I would like to suggest to make this piece of code a function on its own.
```typescript
const getPropertyId = (propertyPath: <types>) => {
  if (isProperty(propertyPath)) {
    const { id } = propertyPath;
    return Array.isArray(id) ? id[id.length - 1] : id;
  }     
    return propertyPath;
}
```
And then change the `let` to `const`
...
```typescript
const propertyId = getPropertyId(propertyPath);
```

RobD-tech: This code is literally copied from `datePickerInput.tsx`. I agree with the suggestions and that the code quality can be improved, but unfortunately I don't have the time to refactor all the code that the product department originally wrote.

PR #3459, opened 2024-09-09
Fix/default value property id bug page 4706

dylanBetty:


ingmar-stipriaan (dismissed):

src/components/multiAutoCompleteInput.js
```diff
@@ -150,6 +150,9 @@
         ) {
           return parsed.map((obj) => String(obj.id));
         }
+        if (typeof parsed === 'number') {
+          return [parsed];
```
ingmar-stipriaan, line 154: Parsed?

jrquak (dismissed):

src/components/multiAutoCompleteInput.js
dylanBetty: agree that this name isn't great. I've changed it to parsedValue in latest commit, so it makes a bit more sense.

jrquak (approved):

ingmar-stipriaan (approved):

PR #3491, opened 2024-10-04
datepicker can interpret default meta-data data

ewoutquax:
fix: datepicker can interpret default meta-data data

And refactor the component with:
- more clear flow
- more descriptive function- and variable names
- minimal use of temporary constants
- clear use of internal and external format of date-values

Loaded values, read from the MetaAPI, is parsed according to the MetaAPI format.
Default values are parsed according to the given / default format.

Since the component doesn't know the origin of the value, we first assume it's the MetaAPI, and then try the default format.


src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,259 @@
...
+        maxDate={convertToValidDate(useText(maxValue))}
+        format={getFormat(typeComponent)}
         fullWidth={fullWidth}
-        onChange={changeHandler}
+        size={size}
+        margin={margin}
+        data-component={dataComponentAttributeValue}
+        ampm={!is24HourClockTime}
+        keyboardIcon={
+          typeComponent === 'time' ? (
```
ingmar-stipriaan, line 259: Maybe move this to a helper as well to clean up the params

src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,259 @@
...
+        required={required}
+        clearable="true"
+        disabled={isDisabled}
+        autoOk={closeOnSelect}
+        variant={variant}
         inputVariant={inputvariant}
-        onInvalid={invalidHandler}
-        onError={errorHandler}
         InputProps={{
           inputProps: {
```
ingmar-stipriaan, line 272: Maybe move this to a helper as well to clean up the params

src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,259 @@
...
-        placeholder={placeholderText}
+        value={selectedDate}
+        autoComplete={autoComplete ? 'on' : 'off'}
+        placeholder={useText(placeholder)}
+        label={!hideLabel && useText(label)}
+        error={errorState}
+        helperText={currentHelperText}
+        disableToolbar={disableToolbar}
+        disablePast={disablePastDates}
+        minDate={convertToValidDate(useText(minValue))}
```
ingmar-stipriaan, line 250: Min date doesn't work when I test it. I expect to see a message before I submit the form.
My option value = 01-10-2024
Runtime value = 01-09-2024

ingmar-stipriaan (changes requested):

ingmar-stipriaan (approved):

jeroenreijs (approved):

PR #3495, opened 2024-10-07
fix(PAGE-4710): Remove unused zone from publish prod files.

stefan-betty:
The nrfem does not exist anymore, and thus the azure blob account is not reachable. This caused the master-pipeline to crash when trying to publish to the zones.


BettyMichel (approved):
Good!!

jrquak (approved):

PR #3498, opened 2024-10-08
Fix/datetime picker show default value page 4719

ewoutquax:
The dates should be checked against the given date-format.
When no min/maxDate has been set, the value `undefined` should be returned, or the MUI-component will disable scrolling to other months

### DIFF
--- a/src/components/dateTimePickerInput.js
+++ b/src/components/dateTimePickerInput.js
@@ -199,8 +199,8 @@
     };
 
     function convertToValidDate(dateText) {
-      if (DateFns.isValid(dateText)) {
-        const parsedValue = DateFns.parse(dateText, dateFormat);
+      if (dateText) {
+        const parsedValue = DateFns.parse(dateText, getFormat(typeComponent));
         if (isValidDate(parsedValue)) {
           return parsedValue;
         }
@@ -208,7 +208,7 @@
         const parsedValueWithSlashes = dateText.replace(/-/g, '/');
         return new Date(parsedValueWithSlashes);
       }
-      return '';
+      return undefined;
     }
 
     let DateTimeComponent;
--- a/src/components/dateTimePickerInput.js
+++ b/src/components/dateTimePickerInput.js
@@ -83,17 +83,11 @@
     const META_API_TIME_FORMAT = 'HH:mm:ss';
 
     function getFormat(typeFormat) {
-      const defaultFormat = {
-        date: 'dd/MM/yyyy',
-        time: 'HH:mm:ss',
-        datetime: 'dd/MM/yyyy HH:mm:ss',
-      };
-      const optionFormat = {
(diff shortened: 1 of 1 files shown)


src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,265 @@
...
+    const [currentHelperText, setHelper] = useState(optionHelperText);
+    const [isDisabled, setIsDisabled] = useState(disabled);
+    const [isFirstValidation, setIsFirstValidation] = useState(true);
+
+    B.defineFunction('Clear', () => setSelectedDate(null));
+    B.defineFunction('Enable', () => setIsDisabled(false));
+    B.defineFunction('Disable', () => setIsDisabled(true));
+
+    const META_API_DATE_FORMAT = 'yyyy-MM-dd';
+    const META_API_TIME_FORMAT = 'HH:mm:ss';
```
M-Ryan92, line 83: MetaAPI is designtime. maybe change it to DataAPI? as that is the api that accepts these values?

src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,265 @@
...
+        time: timeFormat || 'HH:mm:ss',
+        datetime: datetimeFormat || 'dd/MM/yyyy HH:mm:ss',
+      }[typeFormat];
+    }
+
+    function isValidDate(date) {
+      return date instanceof Date && !isNaN(date);
+    }
+
+    // useEffect for 'valueText', trigger on first componentRender (and data-loaded)
```
M-Ryan92, line 97: remove commented code?

src/components/dateTimePickerInput.js
ewoutquax: not accepts, but sends. But still: your point is valid

src/components/dateTimePickerInput.js
ewoutquax: it's not commented code, but an explanation about the reponsibility of the function. Since there is no room for a function-name in this setup.

src/components/dateTimePickerInput.js
```diff
@@ -28,287 +28,265 @@
       validationAfterMaxValue,
       validationInvalidValue,
       margin,
-      helperText = [''],
+      helperText,
```
M-Ryan92, line 31: will useText be able to handle undefined?

stefan-betty: It should, its a helper provided from B-helpers. Handling `undefined` should be part of the helper

stefan-betty (approved):

src/components/dateTimePickerInput.js
ewoutquax: Fields are now named:
DATA_API_DATE_FORMAT
DATA_API_TIME_FORMAT

PR #3528, opened 2024-11-11
feat: added disable future dates

Pandabweer:
- Added option to disable future dates to dateTimePicker & datePicker component
- Disabled future & past dates for timePicker

### DIFF
--- a/src/components/dateTimePickerInput.js
+++ b/src/components/dateTimePickerInput.js
@@ -31,6 +31,7 @@
       helperText,
       disableToolbar,
       disablePastDates,
+      disableFutureDates,
       hideLabel,
       use24HourClockTime: is24HourClockTime,
       label,
@@ -253,6 +254,7 @@
         helperText={currentHelperText}
         disableToolbar={disableToolbar}
         disablePast={disablePastDates}
+        disableFuture={disableFutureDates}
         minDate={convertToValidDate(useText(minValue))}
         maxDate={convertToValidDate(useText(maxValue))}
         format={getFormat(typeComponent)}
--- a/src/prefabs/structures/TimePicker/index.ts
+++ b/src/prefabs/structures/TimePicker/index.ts
@@ -1,0 +1,109 @@
+import { component, PrefabReference } from '@betty-blocks/component-sdk';
+import { updateOption } from '../../../utils';
+import { Configuration } from '../Configuration';
+import { options as defaults } from './options';
+
+export enum DateInputTypes {
+  DATE_TIME = 'datetime',
+  DATE = 'date',
+  TIME = 'time',
+}
+
+export const TimePicker = (
+  config: Configuration,
+  descendants: PrefabReference[] = [],
+) => {
+  const options = { ...(config.options || defaults) };
+  const style = { ...config.style };
+  const ref = config.ref ? { ...config.ref } : undefined;
+  const label = config.label ? config.label : undefined;
+
+  const categories = [
+    {
+      label: 'Validation Options',
+      expanded: false,
+      members: [
+        'required',
+        'validationValueMissing',
+        'validationInvalidValue',
+        'minValue',
+        'validationBeforeMinValue',
+        'maxValue',
+        'validationAfterMaxValue',
+      ],
+    },
+    {
+      label: 'Styling',
+      expanded: false,
+      members: [
+        'hideLabel',
+        'backgroundColor',
+        'backgroundColorPopup',
+        'borderColor',
+        'borderHoverColor',
+        'borderFocusColor',
+        'labelColor',
+        'textColor',
+        'placeholderColor',
+        'helperColor',
+        'errorColor',
+      ],
+    },
+    {
+      label: 'Advanced Options',
+      expanded: false,
+      members: ['dataComponentAttribute'],
+    },
+  ];
+
+  if (config.inputType) {
+    let format;
+    switch (config.inputType) {
+      case DateInputTypes.DATE_TIME:
+        format = 'MM/dd/yyyy HH:mm:ss';
+        break;
+      case DateInputTypes.DATE:
+        format = 'dd/MM/yyyy';
+        break;
+      case DateInputTypes.TIME:
+        format = 'HH:mm:ss';
+        break;
+      default:
+        format = '';
+    }
+
+    const update = {
+      value: format,
+      ...(config.placeholder
+        ? { configuration: { placeholder: config.placeholder } }
+        : {}),
+    };
+
+    const key = `${config.inputType}Format`;
+
+    options[key] = updateOption(options[key], update);
+
+    options.type = updateOption(options.type, { value: config.inputType });
+
+    if (config.inputLabel) {
+      options.label = updateOption(options.label, {
+        value: [config.inputLabel],
+      });
+    }
+  }
+
+  if (config.dataComponentAttribute) {
+    options.dataComponentAttribute = updateOption(
+      options.dataComponentAttribute,
+      {
+        value: [config.dataComponentAttribute],
+      },
+    );
+  }
+
+  return component(
+    'DateTimePickerInput',
+    { options, style, ref, label, optionCategories: categories },
+    descendants,
+  );
+};
--- a/src/prefabs/structures/TimePicker/options/advanced.ts
+++ b/src/prefabs/structures/TimePicker/options/advanced.ts
@@ -1,0 +1,8 @@
+import { showIfTrue, variable } from '@betty-blocks/component-sdk';
+
+export const advanced = {
+  dataComponentAttribute: variable('Test attribute', {
+    value: ['DateTimePicker'],
+    configuration: { condition: showIfTrue('advancedSettings') },
+  }),
+};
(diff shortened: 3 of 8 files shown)


src/components/dateTimePickerInput.js
```diff
@@ -253,6 +254,7 @@
         helperText={currentHelperText}
         disableToolbar={disableToolbar}
         disablePast={disablePastDates}
+        disableFuture={disableFutureDates}
```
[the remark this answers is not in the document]
Pandabweer: That could be a solution, but most of the time it's either the full future or the full past. I also wanted to follow the same style as the past. maxDate is also only if you'd want relative max date, if you'd want the complete future it would still be needed. This could also be applied for the past being too restricted as it is now.

How I see it, it would be an addition to the checkboxes.

src/prefabs/structures/TimePicker/index.ts
```diff
@@ -0,0 +1,109 @@
+import { component, PrefabReference } from '@betty-blocks/component-sdk';
```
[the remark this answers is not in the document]
Pandabweer: > Why is this whole TimePicker structure added?
> 
> The DateTimePicker is capable to be used for: date, datetime and time selection.
> 
> There is also not a DatePicker structure, just a DateTimePicker structure. So I think this TimePicker structure is not necessary, and should be removed.

This is due to it not having disableFuture or the disablePast option. In the current prefab, timepicker has disablePast, but it doesn't apply anything. Due to it not doing anything, it shouldn't be included in the same prefab as datetime or date. The code can be the same because it ignores the flags.

Why this is in the same PR is because adding the checkbox disableFuture also adds it to the timepicker, which also doesn't work in this way.

src/prefabs/structures/TimePicker/options/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      condition: showIf('type', 'EQ', 'time'),
+    },
+  }),
+  dateFormat: text('Format', {
+    value: 'dd-MM-yyyy',
+    configuration: {
+      condition: showIf('type', 'EQ', 'date'),
+    },
+  }),
+  datetimeFormat: text('Format', {
```
ingmar-stipriaan, line 70: Do we need this option in the timepicker? since it says dateTimeFormat

src/prefabs/structures/TimePicker/options/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  helperText: variable('Helper text', { value: [] }),
+
+  timeFormat: text('Format', {
+    value: 'MM/dd/yyyy HH:mm:ss',
+    configuration: {
+      placeholder: 'dd/MM/yyyy HH:mm:ss',
+      condition: showIf('type', 'EQ', 'time'),
+    },
+  }),
+  dateFormat: text('Format', {
```
ingmar-stipriaan, line 64: Do we need this option in the timepicker? since it says dateFormat

src/prefabs/structures/TimePicker/options/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      condition: showIf('type', 'EQ', false),
+    },
+  }),
+
+  autoComplete: toggle('Autocomplete', { value: true }),
+  disabled: toggle('Disabled', { value: false }),
+  placeholder: variable('Placeholder', { value: [] }),
+  helperText: variable('Helper text', { value: [] }),
+
+  timeFormat: text('Format', {
```
ingmar-stipriaan, line 57: Do we need this option in the timepicker? since the value contains MM/dd/yyyy which is a date value

src/prefabs/structures/TimePicker/options/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    value: [''],
+    configuration: { allowFormatting: false, allowPropertyName: true },
+  }),
+  value: variable('Value', {
+    value: [''],
+    configuration: { allowFormatting: false },
+  }),
+
+  type: text('Type', {
+    value: 'datetime',
```
ingmar-stipriaan, line 46: dateTime should be time here right?
Also should we even have a type if its only a time picker?

src/prefabs/structures/TimePicker/options/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    configuration: {
+      condition: showIf('property', 'EQ', ''),
+    },
+  }),
+
+  property: property('Property', {
+    value: '',
+    showInReconfigure: true,
+    configuration: {
+      allowedKinds: ['DATE'],
```
ingmar-stipriaan, line 28: This should be TIME instead of DATE I think?

src/prefabs/structures/TimePicker/options/advanced.ts
```diff
@@ -0,0 +1,8 @@
+import { showIfTrue, variable } from '@betty-blocks/component-sdk';
+
+export const advanced = {
+  dataComponentAttribute: variable('Test attribute', {
+    value: ['DateTimePicker'],
```
ingmar-stipriaan, line 5: This this be TimePicker I guess?

src/prefabs/structures/TimePicker/index.ts
```diff
@@ -0,0 +1,109 @@
...
+      ],
+    },
+    {
+      label: 'Advanced Options',
+      expanded: false,
+      members: ['dataComponentAttribute'],
+    },
+  ];
+
+  if (config.inputType) {
```
ingmar-stipriaan, line 59: All these cases are not necessary for a time picker

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+  }),
+  minValue: variable('Minimum value', {
+    value: [''],
+    configuration: {
+      allowFormatting: false,
+      condition: showIf('type', 'EQ', 'date'),
+    },
+  }),
+  validationBeforeMinValue: variable('Value before minimum message', {
+    value: ['Date should not be before minimal date'],
```
ingmar-stipriaan, line 30: Date validation should be a time validation

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+  }),
+  maxValue: variable('Maximum value', {
+    value: [''],
+    configuration: {
+      allowFormatting: false,
+      condition: showIf('type', 'EQ', 'date'),
+    },
+  }),
+  validationAfterMaxValue: variable('Value after maximum message', {
+    value: ['Date should not be after maximal date'],
```
ingmar-stipriaan, line 43: Date validation should be a time validation

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+    configuration: {
+      condition: showIf('type', 'EQ', 'date'),
```
ingmar-stipriaan, line 45: This should never be the case in a time picker right?

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+    value: ['Date should not be before minimal date'],
+    configuration: {
+      condition: showIf('type', 'EQ', 'date'),
+    },
+  }),
+  maxValue: variable('Maximum value', {
+    value: [''],
+    configuration: {
+      allowFormatting: false,
+      condition: showIf('type', 'EQ', 'date'),
```
ingmar-stipriaan, line 39: This should never be the case in a time picker right?

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+  validationValueMissing: variable('Value required message', {
+    value: ['This field is required'],
+    configuration: {
+      condition: showIfTrue('required'),
+    },
+  }),
+  validationInvalidValue: variable('Value invalid message', {
+    value: ['Invalid date'],
+    configuration: {
+      condition: showIf('type', 'EQ', 'date'),
```
ingmar-stipriaan, line 19: This should never be the case in a time picker right?

src/prefabs/structures/TimePicker/options/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  }),
+  minValue: variable('Minimum value', {
+    value: [''],
+    configuration: {
+      allowFormatting: false,
+      condition: showIf('type', 'EQ', 'date'),
```
ingmar-stipriaan, line 26: This should never be the case in a time picker right?

src/prefabs/structures/TimePicker/index.ts
```diff
@@ -0,0 +1,109 @@
+import { component, PrefabReference } from '@betty-blocks/component-sdk';
+import { updateOption } from '../../../utils';
+import { Configuration } from '../Configuration';
+import { options as defaults } from './options';
+
+export enum DateInputTypes {
```
ingmar-stipriaan, line 6: The input types are not necessary, because the type is determined by the file right?

PR #3545, opened 2024-12-10
Feat/add link to box star 523

JorisPannekeet:


src/components/box.js
```diff
@@ -97,6 +106,21 @@
       </Box>
     );
 
+    const Component =
+      hasLink || hasExternalLink ? (
+        <Link
+          href={hasExternalLink ? linkToExternalText : undefined}
```
stefan-betty, line 112: at this point we already checked if `hasExternalLink` exists, we can just use `linkToExternalText`

PR #3555, opened 2024-12-17
feat: add "Display Rich Text" checkbox option to dynamically show HTML content PFR-945

RobD-tech:
Added a checkbox option "Display Rich Text" to the Alert, similar to the checkbox option in [the Text prefab](https://github.com/bettyblocks/material-ui-component-set/blob/d5d1b3e754567d12816bec7a47fcc15c9f5f7de2/src/prefabs/structures/Text/options/index.ts#L33). With this option, you are able to return rich text from the action, and show it as HTML content in the Alert. For example, to highlight certain words or to include an inline text link.

Link to the PFR ticket: [PFR-945](https://bettyblocks.atlassian.net/browse/PFR-945).

### DIFF
--- a/src/components/alert.js
+++ b/src/components/alert.js
@@ -15,6 +15,7 @@
       bodyText,
       allowTextServerResponse,
       allowTitleServerResponse,
+      useInnerHtml,
       dataComponentAttribute,
     } = options;
     const title = useText(titleText);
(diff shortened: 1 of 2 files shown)


RobD-tech:
@jrquak why has this pull request been closed without any comment?

jeroenreijs:
I am closing this. Doesn't seem necessary anymore. Feel free to open it again when needed.

RobD-tech:
This is still a wanted feature, so I'm re-opening this pull request.

jeroenreijs:
Not a fan of using dangerouslySetInnerHTML, but this is discussed in the team. The result/arguments was/were that we use it in another component as well and you have to enable it as a developer. So we trust you to use it wisely.

PR #3572, opened 2025-01-22
Fix/rich text value not serializing correctly page 4765

ewoutquax:
And also leading- and trailing whitespaces.

Removing is only done between closing and opening of tags: not inside
the tags, so we don't change user-input.

This is done because Slate builds a node-tree of the tags, and gets
confused by enters in between tags

### DIFF
--- a/src/components/richTextInput.js
+++ b/src/components/richTextInput.js
@@ -113,8 +113,6 @@
       floatLabel,
     } = options;
     const isDev = env === 'dev';
-    const optionValue = useText(valueProp);
-    const [currentValue, setCurrentValue] = useState(optionValue);
     const [valueKey, setValueKey] = useState(0);
     const labelText = useText(label);
     const [showDropdown, setShowDropdown] = useState(false);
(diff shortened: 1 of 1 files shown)


ingmar-stipriaan (changes requested):

src/components/richTextInput.js
```diff
@@ -124,6 +122,18 @@
     const helperTextResolved = useText(helperText);
     const { current: labelControlRef } = useRef(generateUUID());
 
+    const sanitizeHTML = (rawString) => {
+      // Replace all the whitespace between tags with nothing. Replacement is done when there is ONLY whitespace
+      return rawString
+        .toString()
+        .trimLeft()
+        .trimRight()
```
dylanBetty, line 130: Can't this just be `trim()`?

src/components/richTextInput.js
```diff
@@ -366,6 +376,11 @@
         return jsx('element', attrs, children);
       }
 
+      // if (!Element.isElementList(children)) {
+      //   const attrs = ELEMENT_TAGS.P(el);
+      //   children = jsx('element', attrs, children);
+      // }
+
```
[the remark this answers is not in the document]
ewoutquax: Block has been removed

src/components/richTextInput.js
ewoutquax: Totally forgot to do this change; Suggestion has been applied

dylanBetty (approved):

ingmar-stipriaan (approved):

jrquak (approved):

PR #3574, opened 2025-01-23
fix: parse default value of Select input to number value TECHSUP-10003

RobD-tech:
Parse the default value of the Select input component to a number. Another value than number (e.g. "null" or "test") will be  parsed to an empty value. This fixes an issue where the onChange is triggered twice, see the TECHSUP ticket.

Link to TECHSUP ticket: [TECHSUP-10003](https://bettyblocks.atlassian.net/browse/TECHSUP-10003)

### DIFF
--- a/src/components/selectInput.js
+++ b/src/components/selectInput.js
@@ -45,12 +45,13 @@
     const [interactionFilter, setInteractionFilter] = useState({});
     const [disabled, setIsDisabled] = useState(initialIsDisabled);
     const mounted = useRef(false);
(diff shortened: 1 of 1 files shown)


jrquak (dismissed):

RobD-tech:
Good one. I'll have to implement an extra check so that it will check if the value can actually be parsed to a number first.

dylanBetty (approved):

jrquak (approved):

PR #3656, opened 2025-03-11
fix: reset the setIsFirstValidation state back to true upon clearing

enriquebos:
This fixes the error when clearing the date input.


enriquebos:
https://github.com/bettyblocks/material-ui-component-set/commit/3ebfea3b975f8d6e9e35e856564c10577065e132

PR #3718, opened 2025-05-15
feat: gitignore lock files

enriquebos:
The lock file keep changing every install, since it's not really needed inside the root having it not there makes it so it doesn't get pushed to the branch. Every npm install generates it's own lockfile.


jrquak:
We need these files for the CI/CD functionality, don't remove them.

PR #3720, opened 2025-05-15
fix: heap-size for lint crash

enriquebos:
If you try to commit without using --no-verify the precommit lint will fail due to not having enough memory.
`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

This extra ENV variable in the lint command adds 6gb extra ram in order for it not to crash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "build-fast": "bb components build --fast",
     "dev": "nodemon",
     "dev:fast": "nodemon --config nodemonfast.json",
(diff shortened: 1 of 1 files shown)


jrquak:
This is an issue on your local machine, changes like this should not be enabled by default.

enriquebos:
How much memory is it using when you run it? When running it without changing anything I still get the same issue.

enriquebos:
The issue was that the .tmp folder was too big

PR #3739, opened 2025-06-19
feat: add display logic to DataTableColumn component PFR-1070

RobD-tech:
For more information, see [PFR-1070](https://bettyblocks.atlassian.net/browse/PFR-1070).

### DIFF
--- a/src/prefabs/updateForm.tsx
+++ b/src/prefabs/updateForm.tsx
@@ -265,7 +265,9 @@ const beforeCreate = ({
               {buttonGroupValue === 'anotherPage' &&
                 'Link from another page to this page, and pass the ID property of the model.'}
               {buttonGroupValue === 'thisPage' &&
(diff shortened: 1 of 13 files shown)


src/prefabs/updateForm.tsx
```diff
@@ -675,6 +683,9 @@ const beforeCreate = ({
             case 'thisPage':
               saveThisPage();
               break;
+            case 'other':
+              save({ ...originalPrefab });
```
dylanBetty, line 687: I'm pretty sure this can just be `save(originalPrefab)`

dylanBetty: Ah yeah I see, in that case we can skip it since it also has nothing to do with this PR

dylanBetty (dismissed):

dylanBetty (dismissed):

dylanBetty (dismissed):

PR #3776, opened 2025-08-06
fix: Media component doesn't reset to default width/height after clearing unit value & Parent inherit TECHSUP-10418 & PAGE-3579

enriquebos:
https://bettyblocks.atlassian.net/browse/TECHSUP-10418
https://bettyblocks.atlassian.net/browse/PAGE-3579

### DIFF
--- a/src/components/media.js
+++ b/src/components/media.js
@@ -21,6 +21,8 @@
       linkToExternal,
       linkType,
       dataComponentAttribute,
+      width,
+      height,
     } = options;
 
     const titleText = useText(title);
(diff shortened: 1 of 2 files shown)


src/components/media.js
```diff
@@ -235,58 +240,56 @@
...
         '& > *': {
           pointerEvents: 'none',
         },
       },
-      hasContent: {
-        width: 'fit-content',
-        height: 'fit-content',
+      root: {
+        width: ({ options: { width } }) =>
+          !width || /^\d+$/.test(width.trim()) ? '100%' : width,
```
dylanBetty, line 249: Maybe move this to a constant to make it clearer what it does, and make it reusable
```typescript
const onlyContainsDigits = /^\d+$/.test(width.trim());
```

src/components/media.js
```diff
@@ -235,58 +240,56 @@
...
           fill: '#666D85',
         },
       },
       media: {
-        width: ({ options: { width } }) => width,
-        height: ({ options: { height } }) => height,
+        width: ({ options: { width } }) =>
+          !width || /^\d+$/.test(width.trim()) ? '100%' : width,
+        height: ({ options: { height } }) =>
+          !height || /^\d+$/.test(height.trim()) ? '500px' : height,
```
dylanBetty, line 290: The function can then also be re-used here

PR #3817, opened 2025-09-10
Fix/prevent scroll for decimal input page 5558

dylanBetty:


jeroenreijs (approved):
I think the changes are fine, but aren't we breaking things when removing the functions for setting the current value? They must have been added there for a reason. They have been added in this MR: https://github.com/bettyblocks/material-ui-component-set/pull/3780. With the corresponding ticket https://bettyblocks.atlassian.net/browse/PAGE-5498. You might have checked that out already. Else I think you should.

